PR #13, opened 2016-07-22
Fix archiver crashing on erts path

### LINKED ISSUE #12: Consider adding runtime_tools to the generated config.exs applications
Runtime tools are important if anyone wants to use most of the awesome tooling erlang gives us for debugging running systems. With the `applications` configuration in releases I feel like we have a perfect place to define this - it doesn't really make sense for the application itself to depend on the `runtime_tools`. I think it would be reasonable to include the application by default.

It's relatively large with 296K, but I don't think it matters that much.

michalmuskala:
Closes #12 

Unfortunately, I didn't know how to add a proper test for this. I tested this with my app locally and it seemed to work correctly.


bitwalker:
Thanks!

Issue #18, opened 2016-07-28
Fails to run release in foreground


brainlid:
I can successfully build a release, start it and use a Phoenix application on my computer. However, when I try to run the "foreground" script it fails. I'm wanting to run the script in a docker container, but for now everything is on my development machine while I try to just get it working. 

```bash
$ cd /home/mark/dev/connect/rel/connect/releases/0.0.1/

$ ./connect.sh foreground
Exec: /home/mark/dev/connect/rel/connect/erts-8.0/bin/erlexec -noshell -noinput +Bd -boot /home/mark/dev/connect/rel/connect/releases/0.0.1/connect -mode embedded -config /home/mark/dev/connect/rel/connect/releases/0.0.1/sys.config -boot_var ERTS_LIB_DIR /home/mark/dev/connect/rel/connect/erts-8.0/../lib -args_file /home/mark/dev/connect/rel/connect/releases/0.0.1/vm.args -env ERL_LIBS /home/mark/dev/connect/rel/connect/lib -pa /home/mark/dev/connect/rel/connect/lib/connect-0.0.1/consolidated -extra -- foreground
Root: /home/mark/dev/connect/rel/connect
trap: SIGUSR1: bad trap
```

As I said, the release is working, I can start it, interact with the application (Phoenix) and stop it. It fails when trying to run in foreground. I can't figure out why. Most of the guides for deploying an elixir release in a docker container recommend running it that way.

brainlid:
This issue is fixed by PR in #19 (https://github.com/bitwalker/distillery/pull/19). I tested it and verified that it fixes the issue and runs "foreground" correctly.

Please review/apply his PR.

shanesveller:
If you are running on a Debian derivative, this is likely because of the default behavior of replacing upstream `sh` with `dash`, which implements `trap` differently. In my environment (Docker using the official `elixir:1.3.1` image), I was able to correct this by reverting to the original `sh`:
```shell
echo 'dash dash/sh boolean false' | debconf-set-selections
dpkg-reconfigure -phigh dash
```

brainlid:
Thanks, this worked for solving my problem.

PR #19, opened 2016-07-29
`SIG` prefix does not work in `sh`

vishnevskiy:
As the title says.

This trap does did not work in sh.


brainlid:
Verified to fix issue #18.

merqlove:
Yeah, it works

shanesveller:
I posted a workaround for Debian/Ubuntu users in the comments on #18, which doesn't require a change to Distillery itself.

brainlid:
Thanks @shanesveller, verified that change works. Added it to my docker file used for deploying the release as follows:

```text
RUN echo 'dash dash/sh boolean false' | debconf-set-selections
RUN dpkg-reconfigure -phigh dash
```

bitwalker:
I'm going to document the fix from #18 rather than merge this. Thanks for the PR though!

PR #23, opened 2016-08-03
Fix escript shell function

minijackson:
A `shift` that doesn't need to be here, I think


bitwalker:
Was this broken? Unless I'm mistaken first shift drops the command name, "escript", the second one drops the escript path, so that the remaining args can be passed with `$@`.

minijackson:
Oops, I guess the nodetools function above and a confusion between $0 / $1 must have got me… Sorry for the wasted time.

PR #27, opened 2016-08-05
Update the version for distillery dep

sanrodari:


bitwalker:
I'm going to fix this manually since it's conflicted. Thanks anyway!

PR #28, opened 2016-08-07
Add command cli from exrm

imranismail:


bitwalker:
Is there any reason to add this with Distillery's support for custom commands? I'm not entirely opposed to adding it, but it does feel fairly redundant.

bitwalker:
@imranismail I'm still not sure if Distillery will be the best place to enforce any single way of handling migrations, so it's an open question as to what the best approach is there. I think I'll merge this for the time being and deprecate it later if we have a better solution, but since you're not likely to be the only person doing migrations this way, I don't want to cause unnecessary pain making the jump from exrm.

imranismail:
@bitwalker can't wait to see, what you are gonna come up with. It's still a bit of a pain right now as I have to write a module that is doing exactly what those mix tasks are doing.

rawsyntax:
@bitwalker I've noticed with `command` there are no hooks (as with `start` or `foreground`). Would it make sense to have hooks around the `command` command?

bitwalker:
@rawsyntax The `command` handler runs in it's own node, with all code loaded, but none of the applications started, so I don't know that it makes much sense to have hooks around it, particularly since it's meant to run custom code to begin with. Do you have a particular use case?

PR #31, opened 2016-08-09
Fix boot hooks config in docs

ImJoeWebb:
The keys are pre_start_hook and post_start_hook in Mix.Releases.Profile.


bitwalker:
Thanks!

PR #32, opened 2016-08-09
keep erl_pipes in release dir

manukall:
this was done in exrm in https://github.com/bitwalker/exrm/pull/109. i just stumbled over this when having multiple apps under different users on the same server. 

only the first app will be able to write to `/tmp/erl_pipes` and all others fail.


bitwalker:
Thanks! the boot script got completely reworked so unfortunately a few of these things got missed :(

Issue #21, opened 2016-08-02
Include Consolidated BEAM Files?


vishnevskiy:
We just did a release and noticed the consolidated files were not included in the archive despite being copied to the release. I couldn't find any code that actually handles this and patched it so we can re-release. I am just wonder if we are doing something wrong or is there an oversight?

https://github.com/hammerandchisel/distillery/commit/fc4b92ce1dc9a03eac85bf49340b97ddd043d130

bitwalker:
I could've sworn I did this, but sure enough, it's not there. Could you open a PR with your changes?

bitwalker:
Closing this since #33 was merged.

PR #33, opened 2016-08-10
Include Consolidated BEAM Files

vishnevskiy:
Adding PR based on #21


bitwalker:
We'll also need to add them to the code path, but I have some additional work going on there so I'll take care of it. Thanks!

PR #36, opened 2016-08-11
Fix plugins only specified in an environment

Gazler:
When specifying a plugin in an environment only (and not a release) the
plugin is prepended to the profile.plugins attribute. However since
this defaults to nil, the result is an improper list in the format:

    [SampleApp.TestPlugin | nil]

This commit fixes it so that the profile.plugins defaults to an empty
list, resulting in:

    [SampleApp.TestPlugin | []]

A test has been added which loads a sample plugin from the rel
directory and verifies that the list contains a list with the plugin.
This test will fail when the list is an improper list.

### DIFF
--- a/lib/mix/lib/releases/models/profile.ex
+++ b/lib/mix/lib/releases/models/profile.ex
@@ -15,7 +15,7 @@ defmodule Mix.Releases.Profile do
     include_src: nil, # boolean
     include_system_libs: nil, # boolean | "path/to/libs"
     strip_debug_info: nil, # boolean
-    plugins: nil, # list of module names
+    plugins: [], # list of module names
     overlay_vars: nil, # keyword list
     overlays: nil, # overlay list
     overrides: nil, # override list [app: app_path]
--- a/lib/mix/lib/releases/assembler.ex
+++ b/lib/mix/lib/releases/assembler.ex
@@ -62,6 +62,7 @@ defmodule Mix.Releases.Assembler do
     profile = Enum.reduce(env_profile, rel_profile, fn {k, v}, acc ->
       case v do
         nil -> acc
+        []  -> nil
         _   -> Map.put(acc, k, v)
       end
     end)
--- a/test/config_test.exs
+++ b/test/config_test.exs
@@ -38,4 +38,18 @@ defmodule ConfigTest do
                      default_release: :default, default_environment: :default} = config
     end
   end
+
+  describe "plugin config" do
+    test "read!" do
+      config = Mix.Project.in_project(:standard_app, @standard_app, fn _mixfile ->
+        Mix.Releases.Config.read!(Path.join([@standard_app, "rel", "config.exs"]))
+      end)
+      assert %Config{environments: %{
+                        dev: %Environment{profile: %Profile{plugins: []}},
+                        prod: %Environment{profile: %Profile{plugins: [SampleApp.ProdPlugin]}}},
+                     releases: %{
+                       standard_app: %Release{profile: %Profile{plugins: [SampleApp.ReleasePlugin]}}}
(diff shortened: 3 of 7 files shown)


lib/mix/lib/releases/models/profile.ex
```diff
@@ -15,7 +15,7 @@ defmodule Mix.Releases.Profile do
     include_src: nil, # boolean
     include_system_libs: nil, # boolean | "path/to/libs"
     strip_debug_info: nil, # boolean
-    plugins: nil, # list of module names
+    plugins: [], # list of module names
```
bitwalker: We need to handle this [here](https://github.com/bitwalker/distillery/blob/master/lib/mix/lib/releases/assembler.ex#L64), since changing this to `[]` means that whether you have plugins defined in the environment or not, it will always override whatever is defined in the release. We probably just need to add a case for `[]` and treat it like we do `nil` (e.g. ignore it).

bitwalker:
I think a test here should have a plugin in both the environment and the release, and then assert that both execute all of their callbacks by checking that the output of `mix release` contains the respective strings.

Gazler:
@bitwalker Good spot, that was an oversight there.

I have changed the logic a bit to merge lists when the value is a list. I am using List.wrap for safety if the value isn't a list in the release, but is in the environment, although this might not be necessary.

The integration test has been added to check the output, however there is no output from the `after_cleanup` hook, so I have left that out of the test.

bitwalker:
@Gazler looks good! I wonder if it's the right thing to do (merging plugins). It runs somewhat counter to the general behaviour of "environment always trumps release". I would much rather people build a list of shared plugins in their config, and append plugins which belong only in the environment. Something like this:

```elixir
common_plugins = [MyApp.PluginA, MyApp.Pluginb]

environment :dev do
  set plugins: [MyApp.PluginDev | common_plugins]
end

release :myapp do
  set plugins: common_plugins
end
```

This was one of the reasons why I wanted the config file to be an Elixir script, since it allows you to use the language to take care of boilerplate, amongst other things. Thoughts?

Gazler:
@bitwalker That makes sense.

It wasn't really clear to me that the environment plugins would replace the release plugins. It makes sense after your example. This makes me wonder, does the `plugin` macro really offer any benefit over calling `set plugins: `?

bitwalker:
The `set plugins: list` approach is best for the case I described above, and while the `plugin` macro accumulates the same list, it is really more sugar than anything else right now. It could be extended in the future to take options which can be passed to the plugin during execution though. Whether such a feature would ever be needed, since plugins already get all of the release configuration, I don't know, but it is at least one case that would be more difficult with just `set plugins: list`.

Gazler:
This has been updated so lists are now treated the same as nil.

lib/mix/lib/releases/models/profile.ex
bitwalker: Forgot to change this back to `nil` ;)

bitwalker:
Missed one change (`Profile` still has the default for `plugins` as `[]` instead of `nil`), if you want I can merge this manually and tag you as the author, just let me know :)

Gazler:
On my phone ATM. Wouldn't setting that to nil result in an improper list
again? I'll check when I get back to my PC.

Gazler:
@bitwalker I can confirm that changing the profile back to `nil` instead of `[]` will error on https://github.com/bitwalker/distillery/blob/048f0fd3085b5cb6c78b87b49b18d531b6ee9510/lib/mix/lib/releases/plugins/plugin.ex#L160 as there are no matches.

bitwalker:
Oh yes, you're right. Sorry!

PR #62, opened 2016-08-25
Generate vm.args with configurable cookie

schnittchen:
Previously, the erlang cookie generated in vm.args was easily guessable, putting installations and developers at risk.

I propose making an appropriate config option mandatory in case the user does not provide a vm.args file.

### DIFF
--- a/lib/mix/lib/releases/assembler.ex
+++ b/lib/mix/lib/releases/assembler.ex
@@ -275,7 +275,7 @@ defmodule Mix.Releases.Assembler do
     bin_dir         = Path.join(release.output_dir, "bin")
     bootloader_path = Path.join(bin_dir, name)
     boot_path       = Path.join(rel_dir, "#{name}.sh")
-    template_params = release.profile.overlay_vars
+    template_params = release.profile.overlay_vars |> sanitize_binfile_template_params
 
     with :ok <- File.mkdir_p(bin_dir),
          :ok <- generate_nodetool(bin_dir),
@@ -290,6 +290,36 @@ defmodule Mix.Releases.Assembler do
          :ok <- generate_sys_config(release, rel_dir),
          :ok <- include_erts(release),
          :ok <- make_boot_script(release, rel_dir), do: :ok
+  end
+
+  defp sanitize_binfile_template_params(params) do
+    release = Keyword.get(params, :release)
+    cookie = release.profile.cookie
+
+    cond do
+      !cookie ->
+        Logger.warn "Attention! You did not provide a cookie for the erlang distribution " <>
+          "protocol inside rel/config.exs. For backwards compatibility, the release name " <>
+          "will be used as cookie, which is potentially a security risk! " <>
+          "Please generate a cookie that is not guessable " <>
+          "by a machine and place it inside rel/config.exs. " <>
+          "This will be an error in a future release."
+
+        release_with_cookie =
+          %{ release |
+            profile: %{ release.profile | cookie: release.name }
+          }
+
+        params
+        |> Keyword.put(:release, release_with_cookie)
+      String.contains?(Atom.to_string(release.profile.cookie), "insecure") ->
+        Logger.warn "Attention! You have an insecure cookie for the erlang distribution " <>
+          "protocol inside rel/config.exs, probably because a secure cookie could not " <>
+          "be automatically generated. Please generate a cookie that is not guessable " <>
+          "by a machine and replace the current cookie with it."
+        params
+      true -> params
+    end
   end
 
   # Generates a relup and .appup for all upgraded applications during upgrade releases
--- a/priv/templates/vm.args.eex
+++ b/priv/templates/vm.args.eex
@@ -2,7 +2,7 @@
 -name <%= release_name %>@127.0.0.1
 
 ## Cookie for distributed erlang
--setcookie <%= release_name %>
+-setcookie <%= release.profile.cookie %>
 
 ## Heartbeat management; auto-restarts VM if it dies or becomes unresponsive
 ## (Disabled by default..use with caution!)
--- a/lib/distillery/tasks/init.ex
+++ b/lib/distillery/tasks/init.ex
@@ -81,33 +81,66 @@ defmodule Mix.Tasks.Release.Init do
           {Keyword.get(mixfile.project, :app), :permanent}
         end)
       end)
-    no_doc? = Keyword.get(opts, :no_doc, false)
     release_per_app? = Keyword.get(opts, :release_per_app, false)
     if release_per_app? do
-      [no_docs: no_doc?,
-       releases: Enum.map(apps, fn {app, start_type} ->
+      [releases: Enum.map(apps, fn {app, start_type} ->
          [release_name: app,
           is_umbrella: false,
           release_applications: [{app, start_type}]]
        end)]
+      ++ get_common_bindings(opts)
     else
       release_name = String.replace(Path.basename(File.cwd!), "-", "_")
-      [no_docs: no_doc?,
-       releases: [
+      [releases: [
          [release_name: String.to_atom(release_name),
           is_umbrella: true,
           release_applications: apps]]]
+      ++ get_common_bindings(opts)
     end
   end
 
   @spec get_standard_bindings(Keyword.t) :: Keyword.t | no_return
   defp get_standard_bindings(opts) do
     app = Keyword.get(Mix.Project.config, :app)
-    no_doc? = Keyword.get(opts, :no_doc, false)
-    [no_docs: no_doc?,
-     releases: [
+    [releases: [
       [release_name: app,
        is_umbrella: false,
        release_applications: [{app, :permanent}]]]]
+    ++ get_common_bindings(opts)
+  end
+
+  @spec get_common_bindings(Keyword.t) :: Keyword.t
+  defp get_common_bindings(opts) do
+    no_doc? = Keyword.get(opts, :no_doc, false)
+    [no_docs: no_doc?,
+     cookie: get_cookie]
+  end
+
+  defp get_cookie do
+    if can_generate_secure_cookie? do
+      generate_secure_cookie
+    else
+      # When the :crypto module is unavailable, rather than generating
+      # a cookie guessable by a computer, produce this obviously
+      # insecure cookie. A warning will be emitted every time
+      # it is used (i.e. when vm.args is being generated with it).
+      :"insecure_cookie_in_distillery_config"
+    end
+  end
+
+  defp can_generate_secure_cookie? do
+    case Code.ensure_loaded(:crypto) do
+      {:module, :crypto} -> true
+      _ -> false
+    end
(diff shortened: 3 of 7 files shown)


schnittchen:
I would like to add

- a meaningful error message
- help for generating a nice random cookie

bitwalker:
We should generate a random cookie if one is not provided and we're not going to default to the release name. There's no need to force users to provide a cookie, especially since that may not be known at release time. My own process is to provide that cookie via environment variables and `REPLACE_OS_VARS=true`.

schnittchen:
Even better.

I realize I need to work on the interpolation into an erlang term. Well, and the failing test.

schnittchen:
If we generate a random cookie, it will be different around invocations, and breaks things like edeliver's ping.
Maybe we should warn the user to provide a cookie.

bitwalker:
I would suggest generating the cookie in `mix release.init`, and warning if it is not set (for instance if someone provides their own `rel/config.exs`).

priv/templates/vm.args.eex
```diff
@@ -2,7 +2,7 @@
 -name <%= release_name %>@127.0.0.1
 
 ## Cookie for distributed erlang
--setcookie <%= release_name %>
+-setcookie <%= hd(:io_lib.format("~p", cookie)) %>
```
bitwalker: This will fail, `io_lib:format` expects the second argument to be a list. We should just use `String.to_atom/1` here, and ensure that the value provided in the config is converted to an atom if it's not one already.

lib/mix/lib/releases/assembler.ex
```diff
@@ -497,6 +505,12 @@ defmodule Mix.Releases.Assembler do
       do: :ok
   end
 
+  defp generate_cookie do
+    # 255 is max. atom length, see http://erlang.org/doc/efficiency_guide/advanced.html
+    # Cookies, at least when read from disk, need to be from the printable ASCII range
+    (1..255)|> Enum.map(fn _ -> Enum.random(?\s..?~) end) |> List.to_atom
```
bitwalker: I think I prefer `:"#{Enum.map(1..64, fn _ -> Enum.random(?!..?~) end)}"` here. A few reasons:

- 64 characters, same as a sha256 hash, any more than that and it's just needlessly excessive, but 64 chars is still easy to read/manage on a single line
- Start the character range at `!` instead of `\s`, because there is no point in having spaces in the generated cookie, and I'm not sure how spaces might mess up parsing of the vm.args file by Erlang.
- Generate the string directly inside an interpolation expression since it does the same thing but is more concise. This one is obviously subjective, but I personally prefer it.

schnittchen:
Thanks for your feedback, Paul. It might be a few days for me to finish this.

priv/templates/vm.args.eex
bitwalker: Oops, I meant `Atom.to_string/1`

bitwalker:
No worries, the help is definitely appreciated :)

lib/mix/lib/releases/assembler.ex
schnittchen: From https://github.com/erlang/otp/blob/3b7a6ffddc819bf305353a593904cea9e932e7dc/lib/kernel/src/auth.erl#L351 it looks like space chars should not be a problem, but I'm totally okay with sacrificing one character, especially since it might be a smoother experience with tooling like editors and such.

With 64 chars from the suggested range, we are at around 418 bits, compared to ~94 bits generated by https://github.com/erlang/otp/blob/3b7a6ffddc819bf305353a593904cea9e932e7dc/lib/kernel/src/auth.erl#L376. However, I would consider using http://erlang.org/doc/man/crypto.html#strong_rand_bytes-1, since the seeding from :rand looks like it has very little entropy (it basically just uses time, which is considered to be feasible for brute-forcing)

bitwalker: RE: spaces, yeah let's omit them from generated cookies, but it's good to know that they are supported in any case.

I'm open to using `crypto:strong_rand_bytes/1`, but I have two things that make me hesitate:

1. It depends on OpenSSL being available, which we can maybe assume will be available, but I'm not sure to what degree that complicates Windows support, if at all. I still want to keep Distillery merge-able into Mix as well, at least in the near term, and I suspect such a dependency would not be desired.
2. The binary generated can contain unprintable characters, which we could probably solve by sha256 hashing the binary, so maybe it's not an issue

I also want to be clear that while I want to generate reasonably secure cookies, nodes should not be remotely accessible via the internet, and I have no plans to protect that use case out of the box - if someone wants to do that, that is their prerogative, but I can't ever guarantee the defaults will be perfectly secure. This cookie should be sufficiently random to prevent collisions (within reason), printable, and keep distillery free of any additional dependencies, and that's about as far as I wish to take any defaults provided.

schnittchen: I don't know about OpenSSL availability on all target platforms, maybe somebody can chime in here and give insights.
Re (2), I'd user `strong_rand_bytes` and then transform to the printable range.

Regarding accessibility from the internet, one point that is often forgotten is a developer's machine. While less of an interesting target in itself, a developer might leave his/her local machine and node/cluster attackable for remote code execution.

So the approach should be for now: as high an entropy we can guarantee, while keeping portability.

schnittchen: http://erlang.org/doc/installation_guide/INSTALL.html talks a bit about openssl. Apparently it is not strictly required for erlang, and `strong_rand_bytes` would be missing then.

I'll make that explicit in the code, and advise people with a key generation snipped using `strong_rand_bytes` assuming most people have that available.

schnittchen:
I started from scratch, to avoid getting lost.

My new approach is to generate a safe cookie if possible (i.e. openssl is linked), and to generate an obviously insecure cookie together with a warning every time it is used otherwise. Generation now happens during `mix release.init`.

What this means for the user:

- most users will be safe by default
- for those we cannot secure automatically, we are not giving a false sensation of secureness, but we are being explicit instead
- those users can choose to ignore the warning, circumvent it (easy) or exchange the cookie for a secure one.

I'm not sure the way the warning is being output is the best.

Aside from that, I noticed problems with the character range, and my assumption on the format of the vm.args file was wrong. There's no documentation out there for it, and I suspect it is being used by more than just the boot script. The cookie value seems not to be an erlang term, but a plain string. The way the boot script extracts it seems to have problems with a few more characters, so I reduced the range to ?+..?~ excluding the notorious backtick character.
This seems to work for me, but let's keep an eye on that.

bitwalker:
Lines in `vm.args` are converted into arguments to the VM when it starts up, so the values it contains must be valid command-line arguments. I'll see if I can figure out why the boot script breaks on weird characters.

schnittchen:
I noticed a couple of "expensive" tests are failing, but at first sight it looks independent of my changes.

schnittchen:
The parsing of the cookie out of `vm.args` happens at two different levels.

One is exlexec: relevant spots are https://github.com/erlang/otp/blob/maint/erts/etc/common/erlexec.c#L1854 and https://github.com/erlang/otp/blob/maint/erts/etc/common/erlexec.c#L1667

The other is https://github.com/bitwalker/distillery/blob/master/priv/templates/boot.eex.

At first sight, the boot script seems to deploy correct quoting. Chars special to the parsing in erlexec seem to be whitespace, single and double quotes and backslash.

I'll account for that and perform some testing.

schnittchen:
Yessss, tests passing :dancer:

schnittchen:
bump

bitwalker:
Looking good :)

The only issue I can think of is that legacy `rel/config.exs` files will not have a cookie set, which means we need to default the cookie in `vm.args` to the release name like we were doing previously so that we do not break existing releases. It occurs to me that maybe the best approach here is to use the release name instead of `:insecure_cookie_in_distillery_config`, and then check to see if the release name is being used as the cookie to determine if it's insecure or not. That way the warning is good for both new projects and legacy projects.

schnittchen:
Currently there's three cases:

- release.init can generate secure cookie
- release.init is unable to generate secure cookie
- rel/config.exs is legacy and is missing cookie

I believe we should always generate a cookie assignment in release.init, even when openssl is not available. In the latter case, we can latch on the legacy behaviour, or insert the stub cookie. 

At some point in the future, there should not be config files without a cookie and we can simply abort when that happens. We should warn both at generation and build time when the stub cookie is inserted or seen.

WDYT?

bitwalker:
I agree with everything you've said, but I think the path I would like to take to make it less painful for current users is to use the release name as the cookie, which means a warning will be produced both for legacy projects and new ones, and we can default to always generating a cookie for new projects via `release.init`. For 1.0, we can force the breaking change, and use the ugly stub cookie by default if we are unable to generate one, and abort if a cookie is not provided at all. Best of both worlds :)

I don't have a specific way of tracking those things currently, other than my own notes. If you'd like to add a TODO file to the project root, and start a list of things we need to do for the transition, I'll start using that to track such things.

schnittchen:
I have all checks and warnings in place now (in another branch), but I just noticed warnings from `Logger.warn` aren't shown to the user when invoking distillery from edeliver. Do you know if there's a better way?

bitwalker:
I suspect edeliver must be overriding how the verbosity levels are handled, but in the default Distillery configuration, warnings are always shown unless the logger is configured to be in silent mode.

schnittchen:
Maybe edeliver captures stdout/stderr and only shows it when --verbose is passed. I guess we have to live with that.

Regarding legacy support and warnings, I think we were both saying the same. Commits have been added.

TODO.md
```diff
@@ -0,0 +1,5 @@
+# For upcoming release
```
bitwalker: Maybe we should just use the new Projects feature for this repo?

bitwalker:
Just the one comment about how to manage TODOs and we can get this merged I think

TODO.md
schnittchen: Uh? Github foo I don't now about? :)

schnittchen: "You don't have permission to create new projects in this repository"

bitwalker: I'll create it :) you just have to remove this file, then squash your commits, and I'll merge

schnittchen: Do you want the entire PR in one commit?

bitwalker: Yes, unless there is a reason to break it up into pieces, but it all seems logically connected to me, so I'd prefer to squash it to a single commit.

schnittchen:
"It works on my machine..." (Tests pass locally on 18.3+1.3.2)

bitwalker:
It's likely just a hiccup, happens every once in awhile due to the complexity of the tests, since it passed on your machine, and the tests passed for 19.0, I'll assume it's good for now, and merge it. I'll follow up with testing before I release it.

bitwalker:
Thanks again for the PR, and your patience ;)

schnittchen:
Thank you Paul, what a nice ride!

PR #67, opened 2016-09-05
only add release dependencies to code paths in command command

manukall:
### Summary of changes

I'm not sure if this is a good solution. We were running into problems before, because ```````$ERTS_LIB_DIR`  included all versions of all dependencies we ever deployed.

### DIFF
--- a/priv/templates/boot.eex
+++ b/priv/templates/boot.eex
@@ -597,10 +597,11 @@ case "$1" in
         # Save extra arguments
         ARGS="$@"
 
+        __code_paths=$(_get_code_paths)
         # Build arguments for erlexec
         set -- "$ERL_OPTS"
         [ "$SYS_CONFIG_PATH" ] && set -- "$@" -config "$SYS_CONFIG_PATH"
-        set -- "$@" -boot_var ERTS_LIB_DIR "$ERTS_LIB_DIR"
+        set -- "$@" "$__code_paths"
         set -- "$@" -noshell
         set -- "$@" -boot start_clean
         set -- "$@" -s "$MODULE" "$FUNCTION"


bitwalker:
Hmm, I'll need to a do a bit of testing with this one, I think we need to set the boot var, but I need to confirm what's using it, maybe it's not needed if we're manually setting the code paths.

manukall:
thanks for looking into this. let me know if i can help.

can you point me to some documentation that might clear that up for me? i'm not sure i understand why libraries that are not part of the applications list should be required in any case. not trying to convince you here, but i'd like to understand and maybe be a little more helpful :-)

bitwalker:
Only libraries which are in the applications list will be present in the release, but the boot var `ERTS_LIB_DIR` is expanded in the `<release_name>.script` file of the release when `include_erts: false`, and the `.script` only loads specific versions of applications for the given release. So we can't remove the boot var. I'd like to figure out why this was causing the issue you describe, because as far as I can tell it's not possible, especially since the `command` task is booting with `start_clean` which loads only the Erlang `kernel`, `stdlib`, and `compiler` applications, as well as `elixir` and `iex`, and specific versions at that.

manukall:
i'm not sure i fully understand this.  we're including erts in our release though.

when we deploy a new release for our app, that release is extracted over the previous version, so when we update some library (e.g. ``  some_lib`    from ` 1.0``  to `1.1`)  ` my_app/lib`   contains both ` some_lib-1.0`  and ` some_lib-1.1` . now `  ERTS_LIB_DIR`   points to ` my_app/lib`  so both versions will be in the load path. the modules from ` 1.1`  win, because of the higher version and we don't have a way of downgrading anymore.

bitwalker:
`ERTS_LIB_DIR` is not blindly added to the code path, the boot var is *only* used for expansion within the `.script` file generated per-release, so only instructions in that file add stuff to the code path, with the exception of tasks in the shell script which use `-pa` to append additional code paths.

manukall:
ah, now i'm getting what you are saying :-) ` ERTS_LIB_DIR` is only used in the ` myapp.script`  file if ` include_erts: false` . that's why i couldn't find it anywhere in my release directory. i think i confused it with ` ERL_LIBS` before. i'll look a bit more into what was going wrong with our release. thanks for your patience.

btw., where does ` ERTS_LIB_DIR`  come from? i couldn't find it anywhere in ther elang manual.

bitwalker:
Precisely :)

Boot vars are user-defined (but some are also predefined, such as `ROOT`), so we're using `ERTS_LIB_DIR` as the name for the var which defines the path to the lib directory for Erlang system libraries, but it could've been something else just as easily.

manukall:
@bitwalker thanks again. the dots connected :-) i also found out about `variables` in `:systools.make_script`. now it alls makes sense. i'll close this, as i don't think it fixes anything :-)

manukall:
@bitwalker i think i understood our problem now: we were naming our release versions based on the current datetime, something like `YYYY-MM-DD-HH-mm `. now as in the `start_clean`  script the application itself is not present, `Application.load(:my_app)` will not load the last released version, but another one it finds in the load path (I believe the one with the highest SemVer version). the same would be true for other applications, of course. 

i'm not really sure how to handle this. on one hand i think it would make sense for a release to reference the explicitly defined versions of apps it should include, on the other hand i see the benefit of "starting clean" for debugging and so on.

do you think i would make sense to at least include the release app (`my_app`) in the `start_clean.rel` for cases like this? probably only load it and not start it?

bitwalker:
Take a look at how the `console_clean` task works in the boot script, I basically do what you want there - all of the code paths for the specific version of the release are added, but no applications are loaded/started by default (other than the bare minimum required), but since no code paths other than the ones specific to the given release version are present, running `Application.load(:myapp)` can only load one possible version of that application.

manukall:
Would it make sense to add this to the `command` task as well? We were using it as suggested in the [plataformatec blog post about migrations with exrm](http://blog.plataformatec.com.br/2016/04/running-migration-in-an-exrm-release/) and I can imagine other people might be surprised as well by the issues we ran into.

Otherwise, would this be a use-case for a "custom command"?

bitwalker:
Yes it definitely makes sense, as `command` should work like `console_clean`, where all applications are on the code path, just not loaded/started. I'll add a TODO

manukall:
thanks! would you want me to help? i guess it wouldn't be just adding `-pa "$CONSOLIDATED_DIR" ${_get_code_paths}`?

bitwalker:
Just pushed a change to master which should resolve it, want to give it a shot?

manukall:
i think there are two tiny mistakes: `set -- "$@" ${_code_paths}` should be `set -- "$@" -pa ${__code_paths}`.
(`-pa` needs to be there and there should be two `_` before `code_paths`.) 

after fixing that it worked for me.

bitwalker:
@manukall `${__code_paths}` is a string of `-pa <path> ..`, so the leading `-pa` is not needed (see the other tasks for examples). I do notice that I put the consolidated directory beams out of order though, so that may have caused a problem.

bitwalker:
Pushed a new commit with those fixes.

manukall:
thanks. all is great now :-)

PR #69, opened 2016-09-06
you can't use '-' in an application name

hykw:
### Summary of changes
application name should be `phoenix_distillery` not `phoenix-distillery`


bitwalker:
Good catch :)

PR #71, opened 2016-09-06
Add a check for erl. Fixes #70

### LINKED ISSUE #70: bin/<app_name> dies silently if erl is not found
### Steps to reproduce

Remove the directory `erl` lives in from `$PATH`. Run `/bin/<name>`. There will be no output.

### Verbose Logs

No output. Exit status of 1.

### Description of issue

I expect to see a message explaining why the command failed.

rickr:
### Summary of changes

When running a release on a system that didn't have the `erl` script in the PATH I got no output (even help). This PR adds a note saying erlang wasn't found and points the user to the area they should look in.

### Checklist

### DIFF
--- a/priv/templates/boot.eex
+++ b/priv/templates/boot.eex
@@ -2,6 +2,7 @@
 
 set -e
 
+command -v erl >/dev/null 2>&1 || { echo "erlang runtime not found. If erlang is installed check your $PATH. aborting.";  exit 1; }
 
 # Set VERBOSE to enable verbose logging of the boot script
 #VERBOSE=true


priv/templates/boot.eex
```diff
@@ -2,6 +2,7 @@
 
 set -e
 
+command -v erl >/dev/null 2>&1 || { echo "erlang runtime not found. If erlang is installed check your $PATH. aborting.";  exit 1; }
```
bitwalker: Hmm, the only place `erl` is pulled from the path is in `_find_erts_dir` on line 115, I think this check should be there instead, since we don't need `erl` to be on the path if we're bundling ERTS.

Also, this command will start erl, but not stop it, leaving a dead VM instance laying around. The command should be `erl -noshell -eval "halt()."`, which will start the VM and immediately stop it without attaching a shell.

rickr: If that's the only place it's pulled from path does it make sense to just bail out if __erl is empty?

bitwalker: Yes definitely :)

PR #75, opened 2016-09-12
make bash script unset CDPATH to avoid no such file or directory errors

mikestok:
### Summary of changes

### Checklist

I'm not quite sure where to put a test for this, or if it's even worth a test

When the generated shell script mentioned at the end of `mix release` is
run with CDPATH set on a system where `/bin/sh` is really a `bash` then
`cd` can echo the directory it's changing to. This makes the setting of
`SCRIPT_DIR` fail:

```
 ==> Release successfully built!
    You can run it in one of the following ways:
      Interactive: rel/la_famiglia/bin/la_famiglia console
      Foreground: rel/la_famiglia/bin/la_famiglia foreground
      Daemon: rel/la_famiglia/bin/la_famiglia start
$ rel/la_famiglia/bin/la_famiglia console
rel/la_famiglia/bin/la_famiglia: line 5: cd: /Users/mike/Projects/influitive/la-famiglia/rel/la_famiglia/bin
/Users/mike/Projects/influitive/la-famiglia/rel/la_famiglia/bin/..: No such file or directory
```

### DIFF
--- a/priv/templates/boot_loader.eex
+++ b/priv/templates/boot_loader.eex
@@ -1,6 +1,8 @@
 #!/bin/sh
 set -e
 
+unset CDPATH
+
 SCRIPT_DIR="$(cd $(dirname "$0") && pwd -P)"
 RELEASE_ROOT_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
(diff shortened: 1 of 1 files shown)


bitwalker:
Looks good to me, I think we can leave this one untested.

PR #79, opened 2016-09-16
boot: Improve release directory resolution

martin-langhoff:
### Summary of changes

More reliable mechanism to resolve the installation directory. Updated to strict shell, tested against dash.

~

Canonicalizes directory, resolving any symlink,
uses (g)readlink -f where available, or falls back
on hand-rolled function.

### DIFF
--- a/priv/templates/boot.eex
+++ b/priv/templates/boot.eex
@@ -2,12 +2,49 @@
 
 set -e
 
-
 # Set VERBOSE to enable verbose logging of the boot script
 #VERBOSE=true
 if [ ! -z $DEBUG_BOOT ]; then
     set -x
 fi;
+
+my_readlink_f() {
+    TARGET_FILE="$1"
+    cd $(dirname "$TARGET_FILE")
+    TARGET_FILE=$(basename "$TARGET_FILE")
+
+    # Iterate down a (possible) chain of symlinks
+    while [ -L "$TARGET_FILE" ]
+    do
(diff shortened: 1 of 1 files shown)


bitwalker:
Unfortunately, we can't rely on bash :(

martin-langhoff:
Can you tell me more? I've done plenty of embedded/distro work, and still this is surprising. Depending on what shells we support, we may be able to get something better than `$0`.

martin-langhoff:
Alright, rolled back the bashisms :-)

bitwalker:
So I went hunting through the relx/exrm issues history, and surprisingly, I couldn't find the original motivation for why we wanted to use `/bin/sh` portable features - the issues that I did find were usually due to the fact that the script was running as `/bin/sh`, which as I'm sure you aware, is a minefield of differences from OS to OS. However, if I recall correctly, the motivation for using `/bin/sh` to begin with was to avoid the requirement on additional dependencies (ie. bash), which I still agree with, though it is a pain in the ass to maintain. That said, if requiring bash, even on embedded environments, unikernels, and the like is not as big of an issue as I think it is, I'd be ok with trying to bash-ify the script. The problem is I'm using a very narrow subset of environments, so I hate to speak for everyone when I make a change here.

martin-langhoff:
It's understandable to be conservative. I've installed `dash` and `checkbashisms` in my dev machine.

PR #80, opened 2016-09-16
boot: re-establish RELEASE_MUTABLE_DIRECTORY, with cleaner semantics

martin-langhoff:
### Summary of changes

Builds on #79 (the patch to review here is 5e3ec09 ) , requires https://github.com/bitwalker/conform/pull/87 .

Brings back R_M_D as `running-enviroment`, to place logs, pipes and generated config files.

* Simplifies symlink-to-dev-files support
* Files in `running-environment` get a comment indicating "Generated, create/edit in path/to/release/foo"
* Does not ovewrite release or user-edited files.
* Supports read-only releases (immutable from the PoV of the process).

### DIFF
--- a/priv/templates/boot.eex
+++ b/priv/templates/boot.eex
@@ -72,7 +72,11 @@ REL_LIB_DIR="$RELEASE_ROOT_DIR/lib"
 ERL_OPTS="<%= release.profile.erl_opts %>"
 # When stdout is piped to a file, this is the directory those files will
 # be stored in. defaults to /log in the release root directory
-RUNNER_LOG_DIR="${RUNNER_LOG_DIR:-$RELEASE_ROOT_DIR/log}"
+
+# RELEASE_MUTABLE_DIR
+RELEASE_MUTABLE_DIR="${RELEASE_MUTABLE_DIR:-$RELEASE_ROOT_DIR/var}"
+
+RUNNER_LOG_DIR="${RUNNER_LOG_DIR:-$RELEASE_MUTABLE_DIR/log}"
 # A string of extra options to pass to erl, here for plugins
 EXTRA_OPTS=""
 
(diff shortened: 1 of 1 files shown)


martin-langhoff:
@bitwalker this is working pretty well with the test umbrella project.

My test is roughly

```
rm -fr rel/master_app 
mix release --verbose
chmod -R ugo+w rel/master_app
RELEASE_MUTABLE_DIR=/tmp/distillery-test rel/master_app/bin/master_app console_clean
```

You can then run `find rel/master_app` and explore the different config files and the injected comments. For example 

```
cat /tmp/distillery-test/sys.config
%%Generated - edit /home/martin/src/exrm-umbrella-test/rel/master_app/master_app.conf or /home/martin/src/exrm-umbrella-test/rel/master_app/master_app.conf/sys.config
[{app_one,[{port,"8090"}]},
 {app_two,[{port,"8091"}]},
 {master_app,[{port,8099}]},
 {sasl,[{errlog_type,error}]}].
```

martin-langhoff:
Rebased on dash-strict version of #79

priv/templates/boot.eex
```diff
@@ -44,7 +72,11 @@ REL_LIB_DIR="$RELEASE_ROOT_DIR/lib"
 ERL_OPTS="<%= release.profile.erl_opts %>"
 # When stdout is piped to a file, this is the directory those files will
 # be stored in. defaults to /log in the release root directory
-RUNNER_LOG_DIR="${RUNNER_LOG_DIR:-$RELEASE_ROOT_DIR/log}"
+
+# RELEASE_MUTABLE_DIR
+RELEASE_MUTABLE_DIR=${RELEASE_MUTABLE_DIR:-$RELEASE_ROOT_DIR/running-environment}
```
bitwalker: Needs to be quoted like `RUNNER_LOG_DIR`

bitwalker: Also I'm kind of in favor of using `tmp` instead of `running-environment`, but I'm open to being convinced if there is some sort of additional significance to the name.

martin-langhoff: Just 'tmp' wouldn't be a good name for a directory that contains the default `RUNNER_LOG_DIR` location.

Quoted, of course, missed that one.

bitwalker: What about `var`? It vibes with your standard UNIX-y directory naming conventions, and `var/log` seems natural to me. Thoughts?

martin-langhoff: Love `var` (that's where we put it, with R_M_D). Updated patch coming in a sec.

martin-langhoff: 9ceefaf for your reviewing pleasure.

bitwalker:
I will merge once the test suite finishes :)

bitwalker:
Thank you!

PR #83, opened 2016-09-20
boot: guard VMARGS_PATH and SYSCONFIG_PATH from recursive calls

martin-langhoff:
... or confused users setting them to the same location where
we expect to write the transient copy.


bitwalker:
Looks good! Good idea :)

PR #86, opened 2016-09-27
Make sure $__erl is not blank

rickr:
### Summary of changes

When running a release on a system that didn't have the `erl` script in the PATH I got no output (even help). This PR adds a note saying erlang wasn't found and points the user to the area they should look in.  

Continuation of #71.

### Checklist


bitwalker:
Looks good! Thanks for the PR!

PR #99, opened 2016-10-05
FIX: extra brace in string

edmz:
### Summary of changes

The extra brace made the search for libraries fail.


bitwalker:
Amazing, I looked over that like 50 times :/

Issue #100, opened 2016-10-05
Distillery should support dynamically creating $HOME/.erlang.cookie in a pre-start hook


mtaylor91:
### Steps to reproduce

  1. Create a pre-start hook to initialize $HOME/.erlang.cookie if not already present:

    ```
    #!/bin/bash

    COOKIE_FILE=${HOME}/.erlang.cookie

    echo "Checking erlang distribution cookie..."
    if ! [ -f "${COOKIE_FILE}" ] ; then
      echo "Initializing secure erlang distribution cookie..."
      dd if=/dev/urandom bs=1 count=32 2>/dev/null | base64 -w 0 > $COOKIE_FILE
      chmod 0400 $COOKIE_FILE
    fi
    ```

### Logs

Release fails to start complaining about distribution cookie not being set:

```
+ printf a secret cookie must be provided in one of the following ways:\n  - with vm.args using the -setcookie parameter,\n  or\n  by writing the cookie to '/home/mtaylor/.erlang.cookie', with permissions set to 0400
a secret cookie must be provided in one of the following ways:
  - with vm.args using the -setcookie parameter,
  or
  by writing the cookie to '/home/mtaylor/.erlang.cookie', with permissions set to 0400+ exit 1
```

### Description of issue

- Sanity check to ensure cookie is set in the generated release script runs before any pre-start hooks.
- The use-case for this is the desire to avoid packaging security-sensitive data (i.e. the distribution cookie) in release tarballs that may or may not always be handled securely.

PR #101, opened 2016-10-05
Defer cookie validation until after pre-start hook(s)

mtaylor91:
This is a change to the boot shell script to address https://github.com/bitwalker/distillery/issues/100

A bit quick and dirty as I didn't have a ton of time to work on this, but the test suite is passing and manual test confirms this has the desired effect of allowing the cookie file to be created in a pre-start hook.

Cheers!
- Mike


bitwalker:
Looks good to me at the moment, I'll give it some time on master before I cut a new release.

Issue #76, opened 2016-09-13
Make hooks into conf.d style dirs [ENH]


martin-langhoff:
It'd be much better to have `start_hook` and friends be _directories_, where scripts found following a certain name pattern are sourced/executed in alphabetical order.

A simple name pattern ("must end with `.hook`, must start with alphanum), helps prevent accidental execution of hidden/temp, conflict files (`.rej`, `.rpmsave`), and other floatsam and jetsam.

It'd be a good idea to do it now (pre `1.0`), as it's API vs other modules.

bitwalker:
I agree, this has been in my mental todo list recently. Let's make this a priority.

martin-langhoff:
I've implemented this a couple times, can bring in known-good code.

bitwalker:
That would be excellent :)

spscream:
Good Idea, but what if project has some dependencies? e.g I have project with conform dependency and conform has its own hook. Now conform hook can be added as pre_start_hook at rel/config.exs. I think it would be good to make possible run hooks from dependencies without any additional configuration, e.g. if dependency has rel/hooks/ directory, then content should be treated as hooks  and must run on release .start/stop/etc.?

spscream:
added implementation ;)

spscream:
@martin-langhoff hm, tried to use hooks implementation with busybox shell inside of alpine:edge docker container, and `read` command returns `read: line 723: illegal option -d`, do you know is there any alternatives to replace `read` command?

martin-langhoff:
Gah, I wasn't expecting ~~the Spanish Inquisition~~ busybox! From what I see, it uses `ash` (or a variant thereof).

From an initial look-around, I don't spot any good alternatives (IFS can't handle null, there's no `ash` mailing list or irc channel to ask ideas, limited docs...). Maybe we can rig something up using `find ... printf "%q"` but I doubt we'll get what we want portably.

What follows is IMHO, and I'm not the maintainer, so feel free to ignore :-)

Aiming for complete lowest common denominator in shell interpreter is, IMO, spending sizable effort for little or no gain. `dash` on F23 x86-64 (optimized for speed, not binary size) is 120k, and that's as bloated as it'll get.

Anyone caring for `dash` vs `busybox` size cannot afford the Erlang runtime footprint.

[...]

PR #102, opened 2016-10-06
Add support for hooks directories

spscream:
### Summary of changes
I have to run multiple hooks for pre_start/post_start/pre_stop/post_stop, but with current configuration it is not possible without hacks. I added new config parameters you can specify directory with hooks located, they will be executed in alphabetical order on release hooks.

This idea was taken from this issue: bitwalker/distillery#76

### Checklist

- [x ] New functions have typespecs, changed functions were updated


bitwalker:
@martin-langhoff mind helping me review this one?

priv/templates/boot.eex
```diff
@@ -208,6 +208,15 @@ _replace_os_vars() {
         }1' < "$1" > "$2"
 }
 
+# Private. Run hooks from directory.
+_run_hooks_from_dir() {
+   if [ -d "$1" ]; then
+          for s in $(find "$1" -name "*.sh" -print | sort); do
```
martin-langhoff: Suggestion:

```
# skip symlinks and other oddities
# handle filenames with escapes and oddities
find "$1" -name "*.sh" -type f -print0 | while read -d $'\0' s ; do
```

I would prefer to inline this grep, to grab only files starting with alphanum, skips '.foo.sh' tmpfiles, but it doesn't work on OSX; the alternative that comes to mind involves ugly bashisms

```
# 
find "$1" -name "*.sh" -type f -print0 | grep -0 '^[0-9a-z-A-Z]' | while read -d $'\0' s ; do
```

spscream: second one isn't working on macosx, I changed to first one. Works for me

martin-langhoff: Works for me.

I did some additional digging; I am used to (defensively) skip any files not starting with alphanum. `vi` and `rsync` used to create `.yourfile.sh` or `.#yourfile.sh`, modern versions of both also add a suffix, so this is less of a concern.

spscream: For me is returning the following:
```
find rel/hooks/pre_start.d/ -name "*.sh" -type f -print0 | grep -0 '^[0-9a-zA-Z]'                                           
Binary file (standard input) matches
```
and it doesn't return a filename

spscream: btw it is incorrect regexp for file path in this case, it won't match in any case, since find is returning a full path to script

martin-langhoff: That's a good point, my grep is erroneous. The `find ... -print0 | while read -d $'\0' s ; do`I think is safe and sane for now.

spscream: `find ./releases/0.0.1/hooks/pre_start.d -name "[0-9a-zA-Z]*.sh"` is working on macosx, alpine linux and ubuntu

martin-langhoff: Excellent! I gave a similar `find` line a try yesteday, and I couldn't make it work on my OSX laptop. Looking now, I had an error in there (opened the name pattern with `^` as if it was a grep pattern) which made me think `find` on OSX couldn't do it.

bitwalker:
Awesome! Thanks for the help @martin-langhoff! Looks good to me as well, so I'll go ahead and merge

spscream:
Thank you guys! distillery is a great package, I'm looking forward to make plugins for it ;)

Issue #96, opened 2016-10-04
0.10.0 breaks REPLACE_OS_VARS


teamon:
Since https://github.com/bitwalker/distillery/compare/0.9.9...0.10.0#diff-8e1a1749b553223a5202b250e93f3e57L180 the `orig_vmargs_path` is no longer set.

### Steps to reproduce

Make a release and set `REPLACE_OS_VARS=true` env variable.

### Verbose Logs

```bash
$ DEBUG_BOOT=true ./myapp.sh
# ...
+ _replace_os_vars  /opt/app/var/vm.args
./myapp.sh: line 1: can't open : no such file
+ awk
        function escape(s) {
            gsub(/&/, "\\\\&", s);
            return s;
        }
        {
            while(match($0,"[$]{[^}]*}")) {
                var=substr($0,RSTART+2,RLENGTH-3);
                gsub("[$]{"var"}", escape(ENVIRON[var]))
            }
        }1
```

spscream:
the same issue with sys.config:
https://github.com/bitwalker/distillery/compare/0.9.9...0.10.0#diff-8e1a1749b553223a5202b250e93f3e57R266

orig_sys_config_path is not set in current version

mtaylor91:
I'm running into this as well; tried using REPLACE_OS_VARS=true as a workaround when my original attempt to dynamically set the cookie using a pre-start hook failed (see https://github.com/bitwalker/distillery/issues/100)

jschniper:
I've run into this as well. Looks like the variable might have been lost in commit 9ceefaf

spscream:
created pr with fix

PR #104, opened 2016-10-06
Fix REPLACE_OS_VARS

spscream:
I fixed work of REPLACE_OS_VARS #96 . 
Since moving of original file is not needed anymore with new mutable directory, I make a awk replacement using temp file.


bitwalker:
Looks good, thanks!

PR #105, opened 2016-10-06
Fix link to upgrades-and-downgrades

edmz:
### Summary of changes

Link was not working because surrounded by backquotes.


bitwalker:
Thanks!

PR #111, opened 2016-10-11
Fixed include_erts

### LINKED ISSUE #108: Match error when setting include_erts and include_system_libs to paths
Hello,

I have a configuration I'd like to use to package an application with the version of `erts` and of the libraries needed by my Raspberry Pi. It looks like the following:

```
environment :raspberry_pi do
  set include_erts: "../erlang_rpi/erts-8.0.1/"
  set include_system_libs: "../erlang_rpi/"
  set include_src: false
end

release :price_watch do
  set version: current_version(:price_watch)
end
```

`erlang_rpi` contains a copy of the libraries I grabbed from the Raspberry Pi installation.

When I try to run it, I get a match error. It looks like setting strings for `include_erts` and `include_system_libs` is triggering a code path that needs apps to be set.

```
** (MatchError) no match of right hand side value: []                                                      
    (distillery) lib/mix/lib/releases/utils.ex:232: anonymous fn/2 in Mix.Releases.Utils.get_apps/1        
    (elixir) lib/enum.ex:1184: Enum."-map/2-lists^map/1-0-"/2                                              
    (distillery) lib/mix/lib/releases/utils.ex:228: Mix.Releases.Utils.get_apps/1                          
    (distillery) lib/mix/lib/releases/assembler.ex:79: Mix.Releases.Assembler.apply_configuration/2        
    (distillery) lib/mix/lib/releases/assembler.ex:27: Mix.Releases.Assembler.assemble/1                   
    (distillery) lib/distillery/tasks/release.ex:108: Mix.Tasks.Release.run/1                              
    (mix) lib/mix/task.ex:296: Mix.Task.run_task/3                                                         
    (mix) lib/mix/cli.ex:58: Mix.CLI.run_task/2
```

Can you please help me figuring out if I'm doing something wrong or if I found a bug?

Thanks,
Mario

mariosangiorgio:
This should fix #108.

Note that this works with the assumptions that the user specifies the path containing both the erts-$VERSION and the lib folders. As I mentioned in the comment to the issue, I am not sure this is the intended path.

If you think the other convention is better let me know and I'll change everything to match with it. For now I changed it this way just because almost all the code was working with this assumption so it was the easiest fix.

Either way, we should document better what the expected value for `include_erts` is and make sure we will fail with the most helpful error message in case there is something wrong. I'd be happy to help for this.

### DIFF
--- a/lib/mix/lib/releases/assembler.ex
+++ b/lib/mix/lib/releases/assembler.ex
@@ -606,7 +606,7 @@ defmodule Mix.Releases.Assembler do
                    case Utils.detect_erts_version(prefix) do
                      {:ok, vsn} ->
                        # verify that the path given was actually the right one
-                       case File.exists?(Path.join(prefix, "bin")) do
+                       case File.exists?(Path.join([prefix, "erts-#{vsn}", "bin"])) do
                          true -> vsn
                          false ->
                            pfx = Path.relative_to_cwd(prefix)
--- a/lib/mix/lib/releases/utils.ex
+++ b/lib/mix/lib/releases/utils.ex
@@ -224,7 +224,7 @@ defmodule Mix.Releases.Utils do
              true  -> apps
              false -> apps
              p when is_binary(p) ->
-               lib_dir = Path.expand(Path.join([p, "..", "lib"]))
+               lib_dir = Path.expand(Path.join([p, "lib"]))
                Enum.reduce(apps, [], fn
                  _, {:error, _} = err ->
                    err


bitwalker:
The path provided to include_erts should always be the prefix, i.e. `/usr/lib/erlang`, not the path to the ERTS folder, i.e. `/usr/lib/erlang/erts-8.0`. Is that what caused your issue?

mariosangiorgio:
Yes, exactly. Also, I see that there are two `bin` directories. One is `erlang/bin`, the other is `erlang/erts-VERSION/bin`. What's the difference between the two?

When I copied things over I copied `erts-VERSION` and `lib`  but I missed `bin`. Then I got confused because the code was failing to pick up `bin` but I had one in `erts-VERSION/bin`.

I think you should ignore this pull request and undo your change (it was based on my broken assumption). Instead, we should validate the user input and check that the path specified has a `bin`, a `erts-VERSION` and a `lib` subdirectory.

Let me know your thoughts on this.

bitwalker:
Agreed, if you want to open a PR for the validation of the `include_erts` setting, I'll be glad to review/merge it. I can tackle it as well if it's not something you have time for. I'll close this PR for now.

Issue #113, opened 2016-10-12
hooks is not working with busybox shell, which is default shell inside of alpine linux


spscream:
If you run release inside of docker alpine:edge container, hooks will not run with error:
```
/app # ./bin/settings start
/app/releases/0.0.1/settings.sh: read: line 723: illegal option -d
/app/releases/0.0.1/settings.sh: read: line 723: illegal option -d
```

it caused by these line:
```
      find "$1" -name "[0-9a-zA-Z]*.sh" -type f -print0 | while read -d $'\0' s; do
```

read on busybox shell doesn't support -d option, so we need to find some alternative solution.

spscream:
@bitwalker, @martin-langhoff, may be we could strip characters from filenames on overlay copy step? so in release hooks dir all filenames with escapes, spaces and other problematic characters will be replaced with '_' char and naive approach without usage of  '\0' line delimiters would work? 
I don't want to install bash in my docker-containers

martin-langhoff:
How 'bout `dash`? :-)

spscream:
I could use any shell, but it will be not a portable solution - someone could use busybox, and after upgrading distillery hooks will not work.
The following works for me in any shell, but I don't know about possible cons of this: 
```
for i in ./hooks/pre_start.d/[0-9a-zA-Z]*.sh; do 
  [ -e "$e" ] || continue
  . $i 
done
```

martin-langhoff:
Works for me (testing across directories with oddly-named files) with minor tidyups... I think we have a winnner.
```
for i in ./hooks/pre_start.d/[0-9a-zA-Z]*.sh; do 
  [ -f "$i" ] || continue
  . "$i" 
done
```

bitwalker:
Looks good, I use alpine in production (via Docker), so I should have some opportunity to see if this causes problems.

PR #114, opened 2016-10-12
Make execution of hooks to work on busybox

spscream:
This pr is fixing bug with execution of hooks inside of busybox shell, bug described in #113


bitwalker:
Thanks!

PR #122, opened 2016-10-19
Fix typo in pre_stop hook overlay

spscream:
Found what pre stop hook overlay copy had  a typo, fixed.


bitwalker:
Good find, thanks for the PR!

PR #125, opened 2016-10-27
Accept --name option as documented for release.init

vic:
--name option was documented but not implemented for release.init task.

Could not find tests for mix tasks suppose only the internal logic is being tested.
Tested it with a local deployment of mine and seemed to work.


bitwalker:
Looks good to me. Thanks for the PR!

PR #130, opened 2016-11-01
Allow passing options to plugins

### LINKED ISSUE #128: Add easy way of configuring plugins
Today, most of the plugins, I've seen are configured by adding stuff to `mix.exs`, but I don't think this is a good option, since distillery has it's configuration infrastructure in `rel/config.exs`.

Unfortunately there's no way of passing options to the plugins. The ideal way would be passing a keyword to the `plugin` - this seems the most natural option.

``` elixir
environment :prod do
  plugin FooPlugin, foo: 1, bar: 2
end
```

The harder part is figuring out how to provide those options to plugins in a backwards-compatible way. Probably the most obvious one would be an additional argument to all the callback functions, but that's not backwards compatible. I propose adding a new field in the `%Release{}` struct called `plugins` that would hold a map of plugin to options configuration - the default would be an empty list. This would mean the options could be accessed with something like:

``` elixir
  def after_package(release) do
    opts = release.plugins[__MODULE__]
    # ...
  end
```

michalmuskala:
### Summary of changes

Closes #128

### Checklist

### DIFF
--- a/docs/Plugins.md
+++ b/docs/Plugins.md
@@ -17,23 +17,29 @@ release :myapp do
 end
 ```
 
+Plugins can be configured by passing options to the `plugin` macro:
+
+```elixir
+plugin MyApp.AwesomePlugin, foo: 1, bar: 2
+```
+
 These plugins are expected to adhere to the `Mix.Releases.Plugin` behaviour.
 
 ### before_assembly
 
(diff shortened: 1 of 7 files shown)


bitwalker:
Top notch work man! Looks good to me.

PR #133, opened 2016-11-05
[TYPO] it's -> its

knewter:
### Summary of changes

Fixes a couple of uses of the word `it's` that were possessive and consequently should have omitted the apostrophe.


bitwalker:
Thanks!

PR #143, opened 2016-11-28
fix nil to atom issue for release.init in umbrella projects when :name is nil

pharosproduction:
### Summary of changes

fix nil to atom issue for release.init in umbrella projects when :name is nil


bitwalker:
Good find!

PR #147, opened 2016-11-29
allow passing a custom eex template to use to the init generator

mobileoverlord:
### Summary of changes

Nerves requires some additional work to be done in the release config. Instead of recreating the generator, we would like to call `release.init` upstream and pass through a path to our custom template.

### Checklist


bitwalker:
Looks good to me!

PR #152, opened 2016-12-01
Update release.init example config to have distinct cookies

### LINKED ISSUE #149: Distinct cookies for dev and prod environments
When running `mix release.init`, a cookie is generated and used in both the `dev` and `prod` environments in `rel/config.exs`.

Would it make sense to instead generate a distinct cookie for `dev` and `prod` each, to help avoid accidentally connecting a `dev` node to the `prod` cluster? Obviously the node names would be the primary way to avoid this, but I still can't think of a reason *not* to use distinct cookies in each environment.

I'd be happy to work on a PR if there's agreement.

tonyvanriet:
### Summary of changes

Closes #149
`mix release.init` will create an example release config with distinct cookies for dev and prod to help avoid accidental node connections between environments.

I debated how to provide the cookies to the config template. I didn't want to remove the existing `:cookie` binding so as to not break any existing custom config templates. Now I see that the custom template feature is brand new, so perhaps that's not a concern.  
I ended up providing an additional binding for the `get_cookie` function so the template could provide cookies for an arbitrary number of environments.  
Let me know if you'd prefer to handle this differently.

I didn't see any existing test coverage for the `release.init` task. I started working on some general `release.init` tests, not specific to this enhancement, and it started to seem like that might deserve it's own PR, so I have that queued up [here](https://github.com/NFIBrokerage/distillery/compare/distinct-cookies...test-release-init). I can merge the two into this PR, if preferred. Otherwise, I'll just submit that PR after integrating this merge.

### Checklist

- [ ] ~~New functions have typespecs, changed functions were updated~~
- [ ] ~~Same for documentation, including moduledocs~~
- [ ] ~~Tests were added or updated to cover changes~~

### DIFF
--- a/lib/distillery/tasks/init.ex
+++ b/lib/distillery/tasks/init.ex
@@ -132,7 +132,8 @@ defmodule Mix.Tasks.Release.Init do
   defp get_common_bindings(opts) do
     no_doc? = Keyword.get(opts, :no_doc, false)
     [no_docs: no_doc?,
-     cookie: get_cookie]
+     cookie: get_cookie,
+     get_cookie: &get_cookie/0]
   end
 
   defp get_cookie do
--- a/priv/templates/example_config.eex
+++ b/priv/templates/example_config.eex
@@ -16,13 +16,13 @@ use Mix.Releases.Config,
 environment :dev do
   set dev_mode: true
   set include_erts: false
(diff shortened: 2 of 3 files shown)


bitwalker:
I think this looks good, let's do the tests in a separate PR. Thanks!

PR #157, opened 2016-12-05
brunch production flag

fregas:
changed the phoenix walkthrough file to use the correct brunch flag for production


bitwalker:
Thanks!

PR #158, opened 2016-12-07
boot: run pre_start hooks when running commands

martin-langhoff:
### Summary of changes

Use case: read the (conform-driven) configuration, to set DB connection, etc for the 'prepare/seed' actions that are the main use case for the `command` facility.


bitwalker:
Looks good!

PR #161, opened 2016-12-09
boot: use vm.args file for `command`

martin-langhoff:
running `command` now gets settings (such as `-name`) from vm.args ...

### DIFF
--- a/priv/templates/boot.eex
+++ b/priv/templates/boot.eex
@@ -723,6 +723,7 @@ case "$1" in
         set -- "$ERL_OPTS"
         [ "$SYS_CONFIG_PATH" ] && set -- "$@" -config "$SYS_CONFIG_PATH"
         set -- "$@" -boot_var ERTS_LIB_DIR "$ERTS_LIB_DIR"
+        set -- "$@" -args_file "$VMARGS_PATH"
         set -- "$@" -noshell
         set -- "$@" -boot start_clean
         set -- "$@" -pa "$CONSOLIDATED_DIR"


bitwalker:
There is a problem with this, in that running a command on the same host as the running application will fail because the node names will be identical. Maybe generating a temp copy with the name/sname settings stripped would be a simple solution?

martin-langhoff:
I have been working under the assumption that if the node is running you
would use `rpc`, and if the node is down (maintenance) you'd use `command`.

For our project we have some related tooling (in Elixir) to expose
functions as Mix Tasks, `rpc`, and `command`, with tags indicating whether
they can be run `cold` (via `command` for offline maintenance) or `hot`
(`rpc` on a running system). Some work either way, some only hot, or only
cold.

If this distinction is wrong, then this patch is wrong too for the general
distillery case. I can probably get the same effect with a hook script.

cheers,


m

michalmuskala:
For me `rpc` is more about things that *must* be evaluated on a running system, while `command` is for things that are not strictly related to the system itself, so don't care if the system is running or not - ecto migrations are a perfect example of that for me.

bitwalker:
My intuition is the same as @michalmuskala, while rpc imposes a requirement on having a running node, command can be used both standalone and with a running system, I think it would be a surprising limitation to not be able to use command when the app is running.

martin-langhoff:
Thinking a bit more about this. 

It seems to me that this concern affects the `-name/-sname` param in `vm.args`. 

Does it make sense to say the various settings that go in `vm.args` /in general/ merit being passed to the erlang runtime for `command` actions, and `-name/-sname` may be a special case?

bitwalker:
I'm not sure that I would say they do, some perhaps, but not all. Many of the types of configuration values that go in vm.args tend to be specific either to the deployment target, the runtime profile of the application, or settings for distribution. While some settings (for example, the node cookie) may be desirable to share, there are many which could perhaps cause unexpected behavior if used as part of the `command` task, just looking through the list [here](http://erlang.org/doc/man/erl.html), I can see quite a few that could cause strange problems.

That said, it would be annoying to have to pass the things you *do* want every time on the command line, so perhaps we need to come up with an alternative means of configuring default VM args for this use case, maybe something like `vm_clean.args` which is used for tasks like `command`. I'm not a huge fan of that, but I'm not sure what other alternatives are available. Somehow identifying the set of things from `vm.args` which can be considered "ok" seems risky/fragile, but would perhaps be more intuitive in the general case.

michalmuskala:
Using `vm.args` for one off commands seems strange to me. You usually set up various system flags there, that actually don't matter for most "command" type things. Maybe we should keep the current behaviour but document how you can easily add a script like `command` but with args? This should be easy with what we have now in distillery.

martin-langhoff:
Happy to discuss more, perhaps in tandem with the other `command` discussions. 

(I'll undo the push of an accidental unrelated commit, apologies).

bitwalker:
I like the idea of pushing this kind of requirement to custom commands as @michalmuskala suggested, though I think it might be desirable to expose a cleaner API from the boot script so that one can easily start a new VM without having to know all the intricate details of fully setting up the VM with configuration, etc. Something like `call <flags> <module> <fun> <args>`, `eval <flags> <script text>`, where the former executes the given MFA, in a VM provisioned with the provided flags, and the latter evals the given script in a VM provisioned with the provided flags. One could still set something up from scratch, but would make a lot of basic scenarios easier to implement.

bitwalker:
I think I'm going to close this for now - I'm happy to discuss improvements to the boot script API, but for now I think this PR can be handled via custom commands.

PR #167, opened 2016-12-14
boot: bring back pingpeer

martin-langhoff:
Use case: we have init script for service/app 'x' which depends on service/app 'y' being up. This can be run single-node or multi-node, so there may be many `y@host` erlang nodes that could be responsive.

Using pingpeer (and a for loop, or the `parallel` utility) we can check whether _any_ of the hosts is serving `y` before starting `x`.


bitwalker:
Looks good :)

PR #176, opened 2017-01-03
Phoenix website is not reachable using https so use http.

omin:
Phoenix website is not reachable using https use http instead.


bitwalker:
Thanks!

PR #179, opened 2017-01-10
fix plugin behaviour and docs

mobileoverlord:
### Summary of changes

Plugin callbacks are being called with arity 1 with the exception of after_cleanup. This fixes the plugin behaviour callbacks and docs

### DIFF
--- a/lib/mix/lib/releases/plugins/plugin.ex
+++ b/lib/mix/lib/releases/plugins/plugin.ex
@@ -9,10 +9,10 @@ defmodule Mix.Releases.Plugin do
   `use Mix.Releases.Plugin`. Then write implentations for the following
   callbacks:
 
-    - `c:before_assembly/2`
-    - `c:after_assembly/2`
-    - `c:before_package/2`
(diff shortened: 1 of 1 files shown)


michalmuskala:
This is not true. The `/2` version is used when available. The `/1` version is used as a fallback and a backwards compatibility measure.

https://github.com/mobileoverlord/distillery/blob/b2729dd9daf34bfae09589eeaaf810bc97928fd0/lib/mix/lib/releases/plugins/plugin.ex#L181-L188

bitwalker:
Ah damn, I knew there was something I was forgetting. I'm going to revert this.

mobileoverlord:
Ah HA! Someone didn't read through the whole module.

mobileoverlord:
this is probably why the default implementations were not made yet. What is opts supposed to contain in the other phases?

bitwalker:
It's a list of options provided via the config, if no options were provided by the user, it defaults to an empty list, or calls the /1 callback if /2 is not implemented. IIRC anyway..

PR #184, opened 2017-01-15
Fix typo in example version number

piisalie:
### Summary of changes

A small documentation change to prevent confusion. If @jeg2 and I are reading it right, it seems like the example version number didn't match the ones from the sentence.


bitwalker:
Thanks!

PR #188, opened 2017-01-17
Reduce the chance of invalid sys.config buring boot

hwatkins:
### Summary of changes

Multiple people getting crash dumps when trying to start applications using distillery and edeliver (https://elixirforum.com/t/edeliver-wont-start-project-at-production/2818).  Traced issue to the way the sys.config file is being created in two steps, leaving a race condition that allows an invalid sys.config file for a small period of time.  Create the output more atomically for a solution. 

### Checklist

### DIFF
--- a/priv/templates/boot.eex
+++ b/priv/templates/boot.eex
@@ -312,10 +312,9 @@ if [ -z "$SYS_CONFIG_PATH" ]; then
     fi
 fi
 if [ "$SYS_CONFIG_PATH" != "$RELEASE_MUTABLE_DIR/sys.config" ]; then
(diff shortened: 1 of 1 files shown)


bitwalker:
Thanks for the PR! Do you know if this PR solves the problem completely?

hwatkins:
I was having the crash dump constantly on startup and after the patch I
haven't seen it, I've updated 3-4 times since then.  Since it's a race
condition it's hard to say with 100% certainty, but my results have been
good.

On Wed, Jan 18, 2017 at 3:13 PM, Paul Schoenfelder <notifications@github.com

bitwalker:
Good to know, thanks again for the contribution!

PR #189, opened 2017-01-18
ignore deps not used in runtime

sasa1977:
### Summary of changes

In Elixir 1.4, we can explicitly state that a dependency is not to be included in `applications` via `runtime: false` option. This PR takes into account this option and ignores such dependencies. Consequently, missing application warnings are not reported for these deps.


bitwalker:
Was just thinking about this yesterday :) thanks for the PR!

PR #192, opened 2017-01-24
Preserve order of :applications in boot file.

f355:
This replicates the behavior of `relx` that we (and likely others) rely on: if the release includes several applications, they are loaded and started exactly in the order in which they are specified in `rel/config.exs` (including their dependencies). This allows applications that come earlier in the list to configure/modify the environment/fiddle in some other way with the ones that come later, as well as their dependencies.

### DIFF
--- a/test/config_test.exs
+++ b/test/config_test.exs
@@ -14,7 +14,7 @@ defmodule ConfigTest do
                         dev: %Environment{profile: %Profile{dev_mode: true, include_erts: false}},
                         prod: %Environment{profile: %Profile{dev_mode: false, include_erts: true, strip_debug_info: false}}},
                       releases: %{
-                        standard_app: %Release{version: "0.0.1", applications: [:elixir, :iex, :sasl, :standard_app]}},
+                        standard_app: %Release{version: "0.0.1", applications: [:elixir, :iex, :sasl]}},
                       default_release: :default,
                       default_environment: :dev,
                      } = config
--- a/lib/mix/lib/releases/config/config.ex
+++ b/lib/mix/lib/releases/config/config.ex
@@ -90,7 +90,7 @@ defmodule Mix.Releases.Config do
       unless is_atom(unquote(name)) do
         raise "release name must be an atom! got #{inspect unquote(name)}"
       end
-      rel = Release.new(unquote(name), unquote(@default_release_version), [unquote(name)])
+      rel = Release.new(unquote(name), unquote(@default_release_version), [])
       Mix.Config.Agent.merge var!(config_agent, Mix.Releases.Config),
         [releases: [{unquote(name), rel}]]
       var!(current_rel, Mix.Releases.Config) = unquote(name)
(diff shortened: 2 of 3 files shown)


test/config_test.exs
```diff
@@ -14,7 +14,7 @@ defmodule ConfigTest do
                         dev: %Environment{profile: %Profile{dev_mode: true, include_erts: false}},
                         prod: %Environment{profile: %Profile{dev_mode: false, include_erts: true, strip_debug_info: false}}},
                       releases: %{
-                        standard_app: %Release{version: "0.0.1", applications: [:elixir, :iex, :sasl, :standard_app]}},
+                        standard_app: %Release{version: "0.0.1", applications: [:elixir, :iex, :sasl]}},
```
bitwalker: Why is `:standard_app` no longer in the applications list? It needs to be, or it won't be included in the release.

bitwalker:
The order in the `.boot` file is worked out by `systools` itself, based on what each application has in it's own applications/included_applications lists. The only time the ordering provided by the user is honored is when `systools` doesn't care which comes first (i.e. there is no explicit dependency which requires one to come before the other).

I think this PR probably requires a test which ensures that an application with complex inter-dependencies does the right thing - I'll need to play around with it before I feel comfortable merging it.

f355:
I should've made it more clear - this applies to `:applications` setting in `rel/config.exs`, NOT the dependencies in `.app` files.
You're right about `systools` working out the order in the `.boot` file, if the order is not obviously broken, it does what you told it to do, and this PR rearranges the order a little.

Consider an example:
- app1 depends on dep1 and dep2
- app2 depends on dep2 and dep3
- app3 depends on dep2
- in config.exs: `applications: [:app1 :app2 :app3]`
- the release is being built for app2

The order of dependencies before this PR would be along these lines (notice that app2 and dep3 get launched before app1):
`[:kernel :stdlib :elixir :iex :sasl :dep1 :dep2 :dep3 :app2 :app1 :app3]`
After the PR:
`[:kernel :stdlib :elixir :iex :sasl :dep1 :dep2 :app1 :dep3 :app2 :app3]`
The order is right, at the moment of app1 startup dep3 and app2 are loaded, but not started, and their configuration could be changed by app1. We actually use this to configure stuff from the environment variables before we start the main application.

test/config_test.exs
f355: It gets added later in `Utils.get_apps/1`, so it's fine.
I'd like to get the rest of "default" applications moved from `config.ex` there as well to keep the entire application resolution thing together, but that's out of the scope of this change.

bitwalker:
It makes sense, I just don't feel that the current test suite goes deep enough to validate that this change doesn't break anything. While the scenario you've described is clear enough, I want to make sure less obvious configurations still do the right thing (and I want to in fact make sure that the scenario you've described works as advertised, it would be a great test to have). Basically, I want to make sure the testing of this change is thorough, since it's minimal on the surface, but potentially breaking for downstream users.

f355:
Fair enough.

surik:
HI! Do you have a plan for this?

f355:
JFYI relx hit the same issue recently, and it got fixed: https://github.com/erlware/relx/pull/606

bitwalker:
Can you rebase on `master` and run `mix format` from the root? I'll take it from there :)

bitwalker:
Actually don't worry about it, I'll take care of it. Thanks!

f355:
Awesome, thank you.

PR #194, opened 2017-01-27
Fix systemd naming typo

develop7:
### Summary of changes

It is not 'System D', it's 'systemd', per https://www.freedesktop.org/wiki/Software/systemd/#spelling

### Checklist

N/A

---
This change is [Reviewable](https://reviewable.io/reviews/bitwalker/distillery/194)


bitwalker:
Thanks!

PR #196, opened 2017-01-31
Corrected the path for '.logo' class

userimack:
### Summary of changes

I have corrected the path given in the doc for "logo" class, given path was  _file: web/static/css/app.css_ but it should be _file: web/static/css/phoenix.css_.


bitwalker:
Thanks :)

PR #202, opened 2017-02-15
Fix win32 executable exit conditions

novaugust:
Was perusing the changes in the codebase for the executables feature and noticed this. Looks like win32 would exit on every run, rather than when run with the `--executable` flag, as was the intention


bitwalker:
/facepalm

Great catch! Thanks for the PR :)

PR #223, opened 2017-03-15
Added run_erl_env parameter to provide runtime configuration of run_erl

gamache:
This patch adds the `run_erl_env` configuration parameter, which
allows the user to apply environment variables to the `run_erl` process
that is responsible for starting and running a release.

Set it like this:

    set run_erl_env: "RUN_ERL_LOG_MAXSIZE=10000000 RUN_ERL_LOG_GENERATIONS=10"

This parameter can be overridden by setting the `RUN_ERL_ENV` parameter
at the command line when running `mix release`.

### Checklist


bitwalker:
Great work! Thanks for the PR!

PR #229, opened 2017-03-20
Pass $@ as a string to escript command in boot template

sivsushruth:
If we just pass $@ to escript, it results in splitting by space
As a result pass "$@" to pass on the exact command line args

**./app escript example.escript '{"id": 1}'** 

used to give 
```
['{"id":', '1}']
```

Now gives, 
```
['{"id": 1}'] 
```

### DIFF
--- a/priv/templates/boot.eex
+++ b/priv/templates/boot.eex
@@ -120,7 +120,7 @@ nodetool() {
     command="$1"; shift
     name=${PEERNAME:-$NAME}
     "$ERTS_DIR/bin/escript" "$ROOTDIR/bin/nodetool" "$NAME_TYPE" "$name" \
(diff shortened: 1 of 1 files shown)


sivsushruth:
@bitwalker this should be terribly easy to review. Let me know if I can help with something else as well. Thanks

bitwalker:
What do you get if you pass multiple arguments? It should break them up properly, but I would like to be sure.

sivsushruth:
It breaks them up properly. Should I add a test ?

bitwalker:
I merged this locally since the build failed (it looks like you rebased on top of a commit I made yesterday which broke the build). Thanks for the PR!

PR #233, opened 2017-03-27
Fixed cookie-generation line in example config

zachgarwood:
### Summary of changes

Apologies for the drive-by PR, but I was banging my head against my keyboard for about an hour trying to emulate the cookie-generation in this example config before I realized I just needed to encode the return value of `:crypto.hash\2` to make it work. Also, I piped the result to `String.to_atom\1` to extend the pipeline, instead of relying on string interpolation.

### Checklist

_(None of the following apply.)_

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -24,7 +24,7 @@ end
 environment :prod do
   set include_erts: true
   set include_src: false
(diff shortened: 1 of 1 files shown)


bitwalker:
Thanks!

PR #249, opened 2017-04-12
Ensure plugins are executed in the order they are defined in config

Gazler:
Ensure plugins are executed in the order they are defined in config
Previously plugins were accumulated using `[new | plugins]` - this would
result in the following case:

    plugin MyPlugin1  # Executed second
    plugin MyPlugin2  # Executed first

It makes more sense to run these in the following order:

    plugin MyPlugin1  # Executed first
    plugin MyPlugin2  # Executed second

This is achieved using `plugins ++ [new]`

This could possibly be a breaking change for anyone relying on reverse
ordering.

### Checklist

### DIFF
--- a/lib/mix/lib/releases/config/config.ex
+++ b/lib/mix/lib/releases/config/config.ex
@@ -176,14 +176,14 @@ defmodule Mix.Releases.Config do
         current_env != nil ->
           env = get_in(config, [:environments, current_env])
           profile = env.profile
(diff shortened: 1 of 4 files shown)


Gazler:
@bitwalker Let me know if you want the docs commit too and I'll squash them down.

bitwalker:
Thanks for the quick PR!

PR #256, opened 2017-04-25
run pre_configure hook before command execution

0xAX:
### Summary of changes

This can be necessary for generation of configuration via conform, for
example for configuration of a database and running migration via

/bin/myapp command MyApp run_migration


bitwalker:
Thanks!

PR #257, opened 2017-04-25
Add templatable signal handler (wip)

rozap:
* set signal_handler key now exists
* added doc for it

Not sure about testing; I don't see tests for similar keys like the pre/post hooks. Suggestions?

### Summary of changes

### Checklist

### DIFF
--- a/priv/templates/boot_loader.eex
+++ b/priv/templates/boot_loader.eex
@@ -16,7 +16,7 @@ _cleanup() {
     trap - HUP INT TERM QUIT EXIT USR1 USR2
     echo ""
     echo "Shutting down.."
-    "$RELEASES_DIR/$REL_VSN/$REL_NAME.sh" rpc init stop
+    "$RELEASES_DIR/$REL_VSN/$REL_NAME.sh" <%= signal_handler %>
     wait $__pid
 }
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -154,3 +154,6 @@ and environments.
         A path to a directory with hooks which will be executed prior to stopping a release
     - post_stop_hooks (path);
         A path to a directory with hooks which will be executed after stopping a release
+    - signal_handler (string)
+        A command which is executed when a HUP INT TERM QUIT EXIT USR1 or USR2 signal is sent
+        to the process. Defaults to `rpc init stop`.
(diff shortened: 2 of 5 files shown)


priv/templates/boot_loader.eex
```diff
@@ -16,7 +16,7 @@ _cleanup() {
     trap - HUP INT TERM QUIT EXIT USR1 USR2
     echo ""
     echo "Shutting down.."
-    "$RELEASES_DIR/$REL_VSN/$REL_NAME.sh" rpc init stop
+    "$RELEASES_DIR/$REL_VSN/$REL_NAME.sh" <%= signal_handler %>
     wait $__pid
```
bitwalker: There's definitely a potential that someone will send a command which doesn't shut down the node, and then end up blocking indefinitely because of the `wait`. We don't have a way to handle different signals in different ways here, and one would not be easy to add, unless we go the route of implementing our own version of the OTP 20 signal handler to tide things over. I'm somewhat leaning towards the latter, because realistically I would like to avoid having Distillery be making any of these decisions, and the migration path to the built-in signal handler will be less painful in the future.

Thoughts?

rozap: imo it's fine if the user of distillery wants to do something potentially stupid like block in a signal handler forever. If the process really needs to be killed, a sigkill can be sent.

We have an situation where we're running elixir in docker containers on mesos (sigh....) and mesos will impose a time window where the container can exit and if it doesn't exit in that window it's forcibly killed. So that's the context where the request for this functionality came from - there's a orchestrator that can kill the erlang vm if it decides it needs to (for better or for worse).

I completely understand if you think this feature will be more painful than useful due to folks using it incorrectly, in which case we can continue using our forked version until OTP 20 when all of this becomes irrelevant :100:

bitwalker: Well I guess what I'm suggesting is that Distillery itself implement a signal handler `gen_event` which can be started as part of the release and subscribed to in order to execute custom code after a given signal - we continue to trap signals, but instead we send those signal events to the signal handler (via RPC). It's a close approximation to how people will be able to use the new OTP 20 stuff, meaning migration when 20 drops will be pretty painless, and all that would be required on Distillery's end is either just removing that stuff after some period of time, or detecting the ERTS version and conditionally enabling/disabling it.

If it's not something you can take on right now, I totally understand, but I feel like it's the right solution for this versus this new setting (which would then need to be maintained and go through deprecation, etc.).

bitwalker:
Now that OTP 20 is released, and the boot script changes are in place to support OTP 20's new signal handling, I'm going to close this. Unfortunately it doesn't solve this problem for <OTP 20, but given the upgrade path to 20 is pretty painless, I don't think it's a good tradeoff in terms of maintenance burden to merge this as-is. Thanks for the efforts though! I hate to turn down PRs, hopefully it won't discourage you from further contributions!

PR #263, opened 2017-05-13
Extend Travid build matrix

samhamilton:
### Summary of changes

Extend the scope of which versions of Elixir and OTP are tested against. Thanks!

### Checklist


bitwalker:
Thanks!

PR #265, opened 2017-05-16
Use USE_ERL_SHELL environment variable to switch iex and erl

ishikawa:
### Summary of changes

I don't want to use `iex` in `bin/app.sh start` because iex's prompt "iex (app@127.0.0.1) 1>" is added at the beginning of each line of the log (I'm using AWS CodeDeploy and EC2). So in this PR,
opted out `iex` if `USE_ERL_SHELL` environment variable is set.

With this modification, I can start application without wired prompts.

```
$ USE_ERL_SHELL=1 ERL_OPTS="-noshell -noinput" bin/app start
```

- Should I write docs and tests? where?

### Checklist

### DIFF
--- a/priv/templates/boot.eex
+++ b/priv/templates/boot.eex
@@ -655,6 +655,13 @@ case "$1" in
         ;;
 
     console|console_clean|console_boot)
+        # By default, iex will be used to startup shell.
+        ENABLE_IEX_OPTS="-user Elixir.IEx.CLI -extra --no-halt +iex"
+
+        if [ ! -z "$USE_ERL_SHELL" ]; then
+            ENABLE_IEX_OPTS=""
+        fi
+
         # .boot file typically just $REL_NAME (ie, the app name)
         # however, for debugging, sometimes start_clean.boot is useful.
         # For e.g. 'setup', one may even want to name another boot script.
@@ -707,8 +714,7 @@ case "$1" in
             -config "$SYS_CONFIG_PATH" \
             -mode "$CODE_LOADING_MODE" \
             ${ERL_OPTS} \
-            -user Elixir.IEx.CLI \
-            -extra --no-halt +iex
+            ${ENABLE_IEX_OPTS}
 
         # Dump environment info for logging purposes
         if [ ! -z "$VERBOSE" ]; then


michalmuskala:
Given the system is perfectly capable of running iex and erl shells at the same time, maybe it would be better to pass this configuration when you start the shell with one of `console`, `remote_console` or `clean_console`?

I could imagine something like `bin/app remote_console --shell erl` with default being `bin/app remote_console --shell iex`

Additionally the iex prompt can be configured, e.g. `config :iex, alive_prompt: ">"` (all options are descrived in `IEx.configure/1`).

ishikawa:
@michalmuskala Thanks for your reply!

>I could imagine something like bin/app remote_console --shell erl with default being bin/app remote_console --shell iex

It looks good, but `remote_console` command already use `USE_ERL_SHELL` environment variable.

>Additionally the iex prompt can be configured, e.g. config :iex, alive_prompt: ">" (all options are described in IEx.configure/1).

At first I tried this option. But I couldn't get it. Even if I configured `config :iex, alive_prompt: ""`, it left **a white space** at the beginning of each line of the log.

michalmuskala:
Wait, the shell prompt is not listed in the logs normally - that's the first time I hear about something like this.

ishikawa:
>Wait, the shell prompt is not listed in the logs normally - that's the first time I hear about something like this.

Yeah, starting application from terminal, the shell prompt is listed in only **the first line** in the logs. But when I deploy application through AWS CodeDeploy and start server from CodeDeploy agent, the shell prompt is listed in every lines in the logs.

NOTE: The log files are `erlang.log.N` of run_erl.

bitwalker:
The use of the `console` task by `start` is because it allows us to use `attach` to attach to the running node's shell directly if necessary (for example, you are unable to connect remotely for some reason). Using any other task will not allow us to do that. Furthermore, you should not be relying on the `run_erl` logs as your primary log persistence layer - rather use a file backend for Logger, and ensure that the `:console` backend is removed in your release environment. The `run_erl` logs have rather confusing rotation semantics (in my opinion), and it is not designed for high-volume logs.

ishikawa:
@bitwalker Thanks for your comment. I'll try to use a file backend and close this PR 😄

Issue #267, opened 2017-05-17
Extending sys.config


walkr:
### Preamble 
I know the release artifact `sys.config`is being generated from `config/config.exs`.
In Erlang, the `sys.config` can contain references to other config files for the purpose of extending the former when the release is started.

For instance, we could have something like this:

```erlang
[{sasl,[{errlog_type,error}]}, "/etc/extra.config", "/etc/extra2.config"].
```

### Reference
----------------------
#### sys.config

[...]

> When traversing the contents of sys.config and a filename is encountered, its contents are read and merged with the result so far. When an application configuration tuple {Application, Env} is found, it is merged with the result so far. Merging means that new parameters are added and existing parameter values overwritten. 
----------------------
[source](http://erlang.org/doc/man/config.html)

### Question

With distillery, is there a way to specify these additional files?

I think this would come extremely handy when deploying to multiple machines, and each could store a local file with some specific configurations.

#### For example

`/etc/extra.config` @ host one

```erlang
[{kernel,
  [
    {sync_nodes_optional, ['two@host', 'three@host']},
    {sync_nodes_timeout, 30000}
  ]}
].
```

`/etc/extra.config` @ host two

```erlang
[{kernel,
  [
    {sync_nodes_optional, ['one@host', 'three@host']},
    {sync_nodes_timeout, 30000}
  ]}
].
```

Cheers!

bitwalker:
It is not directly supported today, but we should be able to add release configuration options to support it, something like `included_configs: ["/etc/extra.config"]`.

bitwalker:
This is already implemented in master

PR #268, opened 2017-05-17
Add support for additional config files to be merged into sys.config

walkr:
### Summary of changes

RE: This request addresses #267 

### Checklist


walkr:
@bitwalker Please review and accept changes if desired!

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 1.5
+
+### Added
+
+- Add new option for environments - `included_configs` - which is
+  a list of paths. Each each path represents an individual config file
```
bitwalker: A couple notes:

- Options in general apply to both environments and individual releases, so I would phrase this "Introduced a new configuration option - ..."
- "Each each" ;)
- Clarify that this option is expecting *runtime* paths, otherwise we are sure to have folks who are expecting configs from their build environment to be included in production, and that's not what this option is for.

docs/Configuration.md
```diff
@@ -106,6 +106,12 @@ and environments.
     - include_system_libs (boolean | string);
         should system libs be included in the release,
         a path to system libs to be included can also be provided
+    - included_configs (path list);
+        The syntax of sys.config allows pointing out to other .config files.
```
bitwalker: I think I'd like to rephrase this:

"Erlang allows `sys.config` to include other `.config` files, Distillery supports this feature via `included_configs`. This option expects a list of paths to include, where each path will be read at runtime when the release boots and the configuration settings it contains will take precedence over the configs that came before it. All such includes will have precedence over the default configuration generated with the release. NOTE: It is recommended to use absolute paths"

lib/mix/lib/releases/assembler.ex
```diff
@@ -583,6 +586,17 @@ defmodule Mix.Releases.Assembler do
     end
   end
 
+  # Extend the config with the paths of additional config files
+  defp append_included_configs(config, []), do: config
+  defp append_included_configs(config, included_configs)
+       when is_list(included_configs) do
+    included_configs = Enum.map(included_configs, &:erlang.binary_to_list/1)
```
bitwalker: I would use `String.to_charlist/1` here instead of `:erlang.binary_to_list/1`

bitwalker:
Looks good! Just a few tweaks and I'll get this merged :)

walkr:
Sorry for the delay! Here are the proposed changes. 

Also, I've noticed tests run super slow now (2.5min). I don't know if it's something on my machine or with the repo, but I thought I'd mention it.

Cheers!

bitwalker:
The tests taking longer is due to the new QuickCheck-style tests I added. I would suggest just running the tests you care about until things look good, then run the whole suite at the end (my approach), as over time the test suite is likely to grow in the amount of time it takes to run, and running the whole suite every time is unfortunately not a great option in the case of a library like distillery where it has to spin up a whole VM and orchestrate it in order to run a single test case :( - there isn't really a way to speed things up too much.

bitwalker:
Thanks for all the work on this!

PR #272, opened 2017-05-28
sys.config and vm.args not being copied correctly during upgrade if they are symlinks

Kabal:
### Summary of changes

Using edeliver and the instructions here: https://github.com/boldpoker/edeliver/wiki/Use-per-host-configuration - sys.config and vm.args become symlinks in the release directory.

I was having trouble getting upgrades to work using this setup and I narrowed it down to where it tries to back up the files and copy the existing ones over.

Using the archive (-a) and force (-f) flags I managed to get upgrades to work using symlinked sys.config and vm.args files.
 
I also added a verbose (-v) flag just to give a bit more information about what is being copied, where.

### DIFF
--- a/priv/templates/boot.eex
+++ b/priv/templates/boot.eex
@@ -613,11 +613,11 @@ case "$1" in
             mv "$REL_DIR/vm.args.bak" "$REL_DIR/vm.args"
         fi
         # Then, backup the packaged configs
-        cp "$REL_DIR/sys.config" "$REL_DIR/sys.config.bak"
-        cp "$REL_DIR/vm.args" "$REL_DIR/vm.args.bak"
+        cp -av "$REL_DIR/sys.config" "$REL_DIR/sys.config.bak"
+        cp -av "$REL_DIR/vm.args" "$REL_DIR/vm.args.bak"
         # Then, substitute in the prepared configs
-        cp "$SYS_CONFIG_PATH" "$REL_DIR/sys.config"
-        cp "$VMARGS_PATH" "$REL_DIR/vm.args"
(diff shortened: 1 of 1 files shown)


priv/templates/boot.eex
```diff
@@ -613,11 +613,11 @@ case "$1" in
             mv "$REL_DIR/vm.args.bak" "$REL_DIR/vm.args"
         fi
         # Then, backup the packaged configs
-        cp "$REL_DIR/sys.config" "$REL_DIR/sys.config.bak"
-        cp "$REL_DIR/vm.args" "$REL_DIR/vm.args.bak"
+        cp -afv "$REL_DIR/sys.config" "$REL_DIR/sys.config.bak"
```
bitwalker: I think what we really want here is `cp --preserve=links ...`, but I could see an argument for `-a` (i.e. `--preserve=all`). I'm not a fan of `-f` because it hides unexpected failures, which we definitely don't want in the case of safely backing up an original file. I also would like to remove the `-v`, as we already have the debugging mode available to show us what's happening in detail if troubleshooting, and I'm of the opinion that verbose output is only useful in exceptional cases, not common ones.

If you could change this from `-afv` to `-a`, I'll get this merged :)

Kabal: Leaving off the -f seems to fail for me. Output from --debug is:

cp: cannot create regular file '/home/web/appname/releases/version/sys.config': Permission denied

But running the command myself I get:

cp: 'path/to/old/sys.config' and '/home/web/appname/releases/version/sys.config' are the same file

So maybe the correct answer is just to test if they are the same file or not before copying?

My bash-fu is not that strong so I am not 100% sure of the best way to do that. Maybe "cmp -s old new"?

Issue #277, opened 2017-06-13
Executable release on OpenBSD


siegfried:
### Steps to reproduce

Run any executable release on OpenBSD.

### Verbose Logs

```
# mix release --executable --verbose
warning: the VM is running with native name encoding of latin1 which may cause Elixir to malfunction as it expects utf8. Please ensure your locale is set to UTF-8 (which can be verified by running "locale" in your shell)
==> Loading configuration..
==> Assembling release..
==> Building release test_project:0.1.0 using environment dev
==> Discovered applications:
  distillery-1.4.0
    from: _build/dev/lib/distillery
    applications:
      :kernel
      :stdlib
      :elixir
    includes: none

  logger-1.4.2
    from: /usr/local/lib/elixir/bin/../lib/logger
    applications:
      :kernel
      :stdlib
      :elixir
    includes: none

  compiler-7.0.3
    from: /usr/local/lib/erlang19/lib/compiler-7.0.3
    applications:
      :kernel
      :stdlib
    includes: none

  elixir-1.4.2
    from: /usr/local/lib/elixir/bin/../lib/elixir
    applications:
      :kernel
      :stdlib
      :compiler
    includes: none

  stdlib-3.2
    from: /usr/local/lib/erlang19/lib/stdlib-3.2
    applications:
      :kernel
    includes: none

  kernel-5.1.1
    from: /usr/local/lib/erlang19/lib/kernel-5.1.1
    applications: none
    includes: none

  test_project-0.1.0
    from: _build/dev/lib/test_project
    applications:
      :kernel
      :stdlib
      :elixir
      :logger
      :distillery
    includes: none

  iex-1.4.2
    from: /usr/local/lib/elixir/bin/../lib/iex
    applications:
      :kernel
      :stdlib
      :elixir
    includes: none

  sasl-3.0.2
    from: /usr/local/lib/erlang19/lib/sasl-3.0.2
    applications:
      :kernel
      :stdlib
    includes: none

  runtime_tools-1.11
    from: /usr/local/lib/erlang19/lib/runtime_tools-1.11
    applications:
      :kernel
      :stdlib
    includes: none

==> Generated overlay vars:
    release_name=:test_project
    release_version="0.1.0"
    is_upgrade=false
    upgrade_from=:latest
    dev_mode=true
    include_erts=false
    include_src=false
    include_system_libs=true
    erl_opts=""
    run_erl_env=""
    erts_vsn="8.2"
[...]

PR #278, opened 2017-06-14
fixed failed extraction on OpenBSD

siegfried:
### Summary of changes

According to https://github.com/bitwalker/distillery/issues/277, current executable releases cannot be run correctly on OpenBSD, and possibly some other systems. This is because `tar` command does not always read standard output as default. In fact, I wonder how many `tar` commands does.

On OpenBSD, `tar` reads the archives from a tape device `/dev/rst0` instead of standard output: http://man.openbsd.org/tar#f
On FreeBSD, `tar` reads the archives from `/dev/sa0` while on Linux it reads from `/dev/st0`: http://man.openbsd.org/FreeBSD-11.0/tar#f
On Mac OS X, it should behave the same as OpenBSD. However, I guess it would use standard output when it is not blank.
```
     -f archive
             Filename where the archive is stored.  Defaults to /dev/rst0.  If
             set to hyphen (`-') standard output is used.  See also the TAPE
             environment variable.
```

In my opinion, using standard output explicitly by specifying a hyphen would be more compatible and stable across all systems.

### Checklist


bitwalker:
Thanks!

PR #286, opened 2017-07-06
Update version to 1.4 and add ‘runtime: false’ in doc deps examples

DavidOliver:
As I understand it, Distillery isn't required at runtime so I added `runtime: false`, as per [the example here](https://elixir-lang.org/blog/2017/01/05/elixir-v1-4-0-released/#application-inference).

### DIFF
--- a/docs/Phoenix Walkthrough.md
+++ b/docs/Phoenix Walkthrough.md
@@ -30,7 +30,7 @@ Next we will install distillery in our `mix.exs`
 ```elixir
   defp deps do
     [ ...,
-     {:distillery, "~> 1.4", runtime: false},
+     {:distillery, "~> MAJ.MIN", runtime: false},
       ...,
     ]
   end
--- a/docs/Phoenix Walkthrough.md
+++ b/docs/Phoenix Walkthrough.md
@@ -30,7 +30,7 @@ Next we will install distillery in our `mix.exs`
 ```elixir
   defp deps do
     [ ...,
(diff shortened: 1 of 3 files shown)


docs/Phoenix Walkthrough.md
```diff
@@ -30,7 +30,7 @@ Next we will install distillery in our `mix.exs`
 ```elixir
   defp deps do
     [ ...,
-     {:distillery, "~> MAJ.MIN"},
+     {:distillery, "~> 1.4", runtime: false},
```
bitwalker: I would prefer to not maintain versions in the individual docs, at some point I'll probably template the versions and add something to the build to generate the docs with proper versions, but at this point let's just use `MAJ.MIN`. Can you modify this and the `Walkthrough.md` change to just add `runtime: false`?

docs/Walkthrough.md
```diff
@@ -21,7 +21,7 @@ Just add the following to your deps list in `mix.exs`:
 
 ```elixir
 defp deps do
-  [{:distillery, "~> 0.9"}]
+  [{:distillery, "~> 1.4", runtime: false}]
```
bitwalker: Could you change the version in this one to `~> MAJ.MIN`

DavidOliver:
Done. I wasn't sure which way to go and naturally picked the wrong one. :)

bitwalker:
No worries :). Thanks again!

PR #287, opened 2017-07-06
In docs, set Phoenix app version with Application instead of Mix in case app is in umbrella

DavidOliver:
[Discussion here](https://elixirforum.com/t/distillery-version-nil-in-sys-config-in-phoenix-umbrella-app/6592).


bitwalker:
Thanks!

PR #288, opened 2017-07-07
Update systemd documentation

clarkema:
Contains two minor updates to the systemd example for `start` usage
(as opposed to 'foreground')

 * The example previously specified both `Type=simple` _and_
   `Type=forking`.  Remove the `simple` version.
 * The prose calls out the importance of including
   `RemainAfterExit=yes`, but this didn't appear in the example
   configuration.

### Summary of changes

### Checklist


bitwalker:
Thanks!

PR #291, opened 2017-07-11
Fixing the Windows boot script

SuperDisk:
### Summary of changes

This fixes the `command` task and allows custom tasks to be created on Windows. The foundations were there, I just fixed a couple bugs (a misplaced carat and a missing flag to `erl`).

The script still barfs a little when calling its `foreground` task but that's for another day.

### Checklist

I think commits can be squashed when merging? I don't know how to do it otherwise...

### DIFF
--- a/lib/mix/lib/releases/assembler.ex
+++ b/lib/mix/lib/releases/assembler.ex
@@ -809,6 +809,9 @@ defmodule Mix.Releases.Assembler do
       {:copy, release.profile.post_upgrade_hooks, "#{hooks_dir}/post_upgrade.d"},
       {:mkdir, "releases/<%= release_version %>/commands"} |
       Enum.map(release.profile.commands, fn {name, path} ->
+        {:copy, path, "releases/<%= release_version %>/commands/#{name}.sh"}
+      end) ++ #Hack. This allows for .bat stuff to be run on windows (instead of just forcing .sh)
+      Enum.map(release.profile.commands, fn {name, path} ->
         {:copy, path, "releases/<%= release_version %>/commands/#{name}.bat"}
       end)
     ] |> Enum.filter(fn {:copy, nil, _} -> false; _ -> true end)
--- a/lib/mix/lib/releases/assembler.ex
+++ b/lib/mix/lib/releases/assembler.ex
@@ -809,7 +809,7 @@ defmodule Mix.Releases.Assembler do
       {:copy, release.profile.post_upgrade_hooks, "#{hooks_dir}/post_upgrade.d"},
       {:mkdir, "releases/<%= release_version %>/commands"} |
       Enum.map(release.profile.commands, fn {name, path} ->
-        {:copy, path, "releases/<%= release_version %>/commands/#{name}.bat"} #HACK
+        {:copy, path, "releases/<%= release_version %>/commands/#{name}.bat"}
       end)
     ] |> Enum.filter(fn {:copy, nil, _} -> false; _ -> true end)
 
--- a/lib/mix/lib/releases/assembler.ex
+++ b/lib/mix/lib/releases/assembler.ex
@@ -809,7 +809,7 @@ defmodule Mix.Releases.Assembler do
       {:copy, release.profile.post_upgrade_hooks, "#{hooks_dir}/post_upgrade.d"},
       {:mkdir, "releases/<%= release_version %>/commands"} |
       Enum.map(release.profile.commands, fn {name, path} ->
-        {:copy, path, "releases/<%= release_version %>/commands/#{name}.sh"}
+        {:copy, path, "releases/<%= release_version %>/commands/#{name}.bat"} #HACK
       end)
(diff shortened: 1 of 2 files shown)


lib/mix/lib/releases/assembler.ex
```diff
@@ -810,6 +810,9 @@ defmodule Mix.Releases.Assembler do
       {:mkdir, "releases/<%= release_version %>/commands"} |
       Enum.map(release.profile.commands, fn {name, path} ->
         {:copy, path, "releases/<%= release_version %>/commands/#{name}.sh"}
+      end) ++ #Hack. This allows for .bat stuff to be run on windows (instead of just forcing .sh)
```
bitwalker: We should rethink how we implement this - things will explode if shell scripts are invoked as `.bat` (and vice versa). I think the best approach is to instead of forcing the extension, require paths to include the extension, and then in the platform-specific scripts, only invoke those extensions which we know we can invoke on that platform (i.e. only invoke `.bat` or `.ps` on Windows).

bitwalker: To remain backwards compatible though, we should give anything without an extension, the `.sh` extension. We can do a deprecation cycle later to always require one.

priv/templates/boot_win.eex
```diff
@@ -2,6 +2,8 @@
 ::
 :: Check the :usage section for information about what this script provides
 
+@echo off
```
bitwalker: Could you explain why you added this? I'm only asking because it's been awhile since I wrote the batch script, and I want to make sure I understand the changes being made.

priv/templates/boot_win.eex
```diff
@@ -434,15 +438,22 @@ if exists "%rel_dir%\vm.args.bak" do (
 
 :: Execute a command in a clean Erlang VM
 :command
-@for /f "tokens=1-2*" %%a in ("%args%") do (
-  @set module=%%a
-  @set function=%%b
-  @set command_args=%%c
-)
-@%erl% -boot_var ERTS_LIB_DIR "%erts_dir%\..\lib" ^
+rem @for /f "tokens=1-2*" %%a in ("%args%") do (
```
bitwalker: Why not remove this rather than comment it?

priv/templates/boot_win.eex
```diff
@@ -493,12 +504,12 @@ if exists "%rel_dir%\vm.args.bak" do (
 @echo Extra Opts          %erl_opts%
 @echo.
 @echo Hooks:
-@for %%i in (%hook_dir%\*.d\*.bat) do (
```
bitwalker: Hooks are all in a `hook_name.d` directory under the main hook directory. I think this change will break those hooks.

priv/templates/boot_win.eex
```diff
@@ -493,12 +504,12 @@ if exists "%rel_dir%\vm.args.bak" do (
...
+@for %%i in (%hook_dir%\*.bat) do (
   echo %%i
 )
 @echo.
 @echo Commands:
-@for %%i in (%command_dir%\*.d\*.bat) do (
+@for %%i in (%command_dir%\*.bat) do (
```
bitwalker: Good catch on this though!

priv/templates/boot_win.eex
SuperDisk: You're right, I didn't use hooks so I didn't notice I broke it (I just changed it to match the commands one as that one didn't work)

priv/templates/boot_win.eex
SuperDisk: In the script practically every line starts with @, which causes the line to not be output on the terminal. This line makes it the default behavior so all the @ aren't necessary anymore.

priv/templates/boot_win.eex
SuperDisk: I dunno, I'll remove this

JonRowe:
We're also working with a Windows deployment environment but we're having an issue with mutable data, see #322

lib/mix/lib/releases/assembler.ex
bitwalker: @SuperDisk have you had a chance to look at this?

SuperDisk: @bitwalker I'll come check this out later tonight if I can.

Kabie:
Found another:
https://github.com/bitwalker/distillery/blob/master/priv/templates/boot_win.eex#L80-L95

The `args` is not set during the loop, so MFA call won't work.

bitwalker:
Is anyone watching interested in taking over this PR so we can get the fixes merged? My Windows env is hosed at the moment

Kabie:
@bitwalker I already made some changes here:
https://github.com/Kabie/distillery/commit/6229b608a4ffdd6cc3764b13d9bf4c4e93a06d48

If it looks OK, I will send a PR.

bitwalker:
@Kabie Those changes look good, feel free to open a PR and I'll get them merged!

PR #304, opened 2017-08-10
Fix the command.sh script

evax:
- Escape the module atom
- Ensure the target module's code is loaded before checking the export
- Ensure CONSOLIDATED_DIR is on the code path
- Fix missing newline escape


bitwalker:
Thanks!

PR #307, opened 2017-08-15
Fix consolidation path

whatyouhide:
In Elixir 1.5, the consolidation path changed. This affected Distillery since Distillery doesn't use `Mix.Project.consolidation_path/1` but used `Mix.Project.build_path/1` and built the consolidation path manually.

This resulted in not consolidated protocols on Elixir 1.5.

I didn't add tests because I didn't find tests for the old behaviour on a quick look and this is a really bad/urgent bug so I'd rather bring it to attention as soon as possible. I verified the patch works for one of our production application which was affected by the previous behaviour.

I'll go ahead and ping you @bitwalker in case you only have mention notifications on :)

PS `Mix.Project.consolidation_path/1` was available on 1.3 (didn't check when it was introduced, might be there since 1.0) so no backwards compat concerns.

### DIFF
--- a/lib/mix/lib/releases/assembler.ex
+++ b/lib/mix/lib/releases/assembler.ex
@@ -63,10 +63,9 @@ defmodule Mix.Releases.Assembler do
         copy_app(app, release)
       end
       # Copy consolidated .beams
-      build_path = Mix.Project.build_path(Mix.Project.config)
+      consolidated_src = Mix.Project.consolidation_path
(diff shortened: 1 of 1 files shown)


josevalim:
Btw, this should be released asap as everyone deploying with Elixir v1.5.1 and distillery is running without consolidated protocols.

bitwalker:
Thanks guys, I'll get this released ASAP

Issue #310, opened 2017-08-22
RemainAfterExit recommendation in Systemd docs


brucepom:
In the [systemd example](https://github.com/bitwalker/distillery/blob/master/docs/Use%20With%20systemd.md) in the docs. There is a recommendation:
 

> It's important that you have RemainAfterExit=yes set, or you will get an error trying to start the service.

My experience using this example on Ubuntu was that I didn't run into any problems when omitting `RemainAfterExit=yes` from the unit file. And that including it prevented systemd from restarting the app if it crashed. I wonder if this should be the recommendation. Perhaps we include a note saying that "if you run into issues starting the service you might try adding `RemainAfterExit=yes` to your unit file. Keep in mind that adding this will prevent systemd restarting your app if it crashes. If you think this makes sense I'm happy to submit a PR. Thanks!

bitwalker:
I wasn't the author of the systemd docs unfortunately, so I haven't verified that the instructions are necessarily correct, especially since some people run `foreground`, others use the daemon via `start`. I have some changes planned in Distillery which should make integration with systemd easier though, namely the creation of a pid file which can be given to systemd, which should be much more reliable than hoping the pid detection is picking up the right pid. Basically, one would then use `Type=forking`, `GuessMainPID=false`, and `PIDFile=<deploy_root/var/myapp.pid` in addition to any other configuration in the unit file.

If you would like to provide a PR which revamps this documentation (assuming you've followed the instructions from beginning to end yourself with success), that would be awesome! It may be worth figuring out what the unit file should be for both `foreground` and `start` use cases, since I believe they need to be different, since `start` starts a daemon and returns, whereas `foreground` is much easier (IIRC, `GuessMainPID` does the right thing there).

OvermindDL1:
I think he created this because of my thread at:  https://elixirforum.com/t/pidfile/7620/2
So you may want to see it for details and how to properly set up the BEAM VM to work with systemd so it will actually restart properly.  :-)

And no, in my experience `GuessMainPID` often fails when using the BEAM VM (at least on my systems).

> I have some changes planned in Distillery which should make integration with systemd easier though, namely the creation of a pid file which can be given to systemd, which should be much more reliable than hoping the pid detection is picking up the right pid.

Awesome, that'd make my PIDFile library obsolete then.  It's whole purpose is just to manage a PID file for a BEAM VM.  :-)

bitwalker:
[...]

PR #316, opened 2017-08-25
Update systemd docs

brucepom:
### Summary of changes

Updated examples for use with SystemD. As discussed in https://github.com/bitwalker/distillery/issues/310.

### Checklist

- Remove the recommendation to set `RemainAfterExit=yes`. This doesn't seem to be necessary and prevents SystemD from restarting your app if it crashes.
- We already had examples for
  - running app in foreground mode
  - running it as a daemon having SystemD attempt to guess the pid
I've added a third example configuration for using SystemD with a pidfile, if you don't want to use foreground and want your app restarted after crashes this is the way to go.

- Add notes about the strengths and weaknesses of the three approaches. I wanted to make it clearer why one would choose a particular configuration over another. The effective differences between the three approaches are
  - how logging is handled
  - whether or not you need to manage a pidfile
  - whether or not systemd can detect and restart your app if it crashes
Hopefully I've explained this in the notes prior to each example.

### DIFF
--- a/docs/Use With systemd.md
+++ b/docs/Use With systemd.md
@@ -40,7 +40,7 @@ Here are three general approaches to running a Distillery release with Systemd:
 		After=network.target
 
 		[Service]
-		Type=*forking*
+		Type=forking
 		User=appuser
 		Group=appuser
 		WorkingDirectory=/home/appuser/myapp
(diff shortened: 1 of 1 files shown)


bitwalker:
Thanks!

PR #317, opened 2017-08-31
Bash: Use correct shebangs

hrefhref:
### Summary of changes

Bash is not always installed at `/bin/bash`. This breaks distillery on FreeBSD for example. Use `/usr/bin/env bash` instead.

No need to change the boot_check script as it's already using `which` to check if bash is present.

### DIFF
--- a/priv/libexec/commands/attach.sh
+++ b/priv/libexec/commands/attach.sh
@@ -1,4 +1,6 @@
-#!/usr/bin/env bash --posix
+#!/usr/bin/env bash
+
+set -o posix
 
 # Attaches to the release console as started via `start`
 # NOTE: This is not a remote console, it's the shell of the running node,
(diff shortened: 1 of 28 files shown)


bitwalker:
The build is failing because I don't think you can pass flags to bash with `/usr/bin/env`. I'm guessing we can just use `set` or something to flip the POSIX bit without having to make it a flag though. If I'm mistaken about that, my bad, but it's definitely broken in this state, so we should fix it however appropriate :)

hrefhref:
Yep @bitwalker you're right :) I didn't knew this one. I've pushed the fix (with `set`).

The tests are still failing locally but I suspect another bsd issue not related to this one.

hrefhref:
Hmm, i'm not sure why it failed this time. It only failed on OTP 20 and with a weird error:

```
  1) property generated cookies are always valid (Distillery.Cookies.Test)
     test/distillery_cookies_test.exs:6
     ** (ExUnit.TimeoutError) property timed out after 180000ms. You can change the timeout
```

Not sure if this error is related to my changes or not.

bitwalker:
Yeah that test is a property-based test and sometimes the generator takes too long to test, I should really set a better timeout on that to avoid the random fails

PR #318, opened 2017-08-31
Fallback to ANSII codes if tput fails

hrefhref:
### Summary of changes

On FreeBSD, tput is using termcap instead of terminfo -- so the codes are not the same. Also it seems that termcap don't have codes for colors.

This fixes uses ANSII codes when tput fails.

### DIFF
--- a/priv/libexec/commands/attach.sh
+++ b/priv/libexec/commands/attach.sh
@@ -1,4 +1,6 @@
-#!/usr/bin/env bash --posix
+#!/usr/bin/env bash
+
+set -o posix
 
 # Attaches to the release console as started via `start`
 # NOTE: This is not a remote console, it's the shell of the running node,
--- a/priv/libexec/commands/attach.sh
+++ b/priv/libexec/commands/attach.sh
@@ -1,4 +1,4 @@
-#!/bin/bash --posix
+#!/usr/bin/env bash --posix
 
 # Attaches to the release console as started via `start`
 # NOTE: This is not a remote console, it's the shell of the running node,
--- a/priv/libexec/commands/command.sh
+++ b/priv/libexec/commands/command.sh
@@ -1,4 +1,6 @@
-#!/usr/bin/env bash --posix
+#!/usr/bin/env bash
+
+set -o posix
 
 # Execute as command-line utility
 #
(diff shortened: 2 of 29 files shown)


OvermindDL1:
> On FreeBSD, tput is using termcap instead of terminfo -- so the codes are not the same. Also it seems that termcap don't have codes for colors.

Wait what?  Back on FreeBSD 4 I used `tput` excessively for getting color codes, is that broken on modern systems?!  o.O

hrefhref:
@OvermindDL1 Oh? Interesting! How are you doing it ?

`setaf` and `setf` don't work for me. Maybe you used the port's/ncurses's tput, and not the system one?

OvermindDL1:
@hrefhref Funny enough I still have my meta-library script from those old days (ahhh ksh88, those days sucked ^.^).  Unsure what version of tput though, but here is a snippet from that *monstrous* library script:
```sh
        typeset RES BLUE WHITE RED GREEN YELLOW 
        RES=`tput rmso`
        BLUE=`tput bold``tput setf 6`
        WHITE=`tput bold``tput setf 7`
        RED=`tput bold``tput setf 1`
        GREEN=`tput bold``tput setf 2`
        YELLOW=`tput bold``tput setf 3`
```

I do remember a lot of things did not match the documentation, like at all, but a lot of testing back and forth got it all figured out on that old system.

hrefhref:
@OvermindDL1 I think you were using the port's tput and not the system one then :) All of theses are failing with system's tput:

```
@ummagumma ~ $ tput rmso
@ummagumma ~ $ echo $?
1

@ummagumma ~ $ tput setf 6
@ummagumma ~ $ echo $?
1

@ummagumma ~ $ tput bold
@ummagumma ~ $ echo $?
1
```

OvermindDL1:
Could make the fixed `tput` a requirement then otherwise no coloring?  ^.^

hrefhref:
I don't like the idea of having another dependency. Falling back to ANSII codes seems the correct way to do it.

And if you have the correct tput, it'll work anyway!

deadtrickster:
please merge it already

deadtrickster:
Hope this will be pushed to hex soon :-). Thanks

PR #322, opened 2017-09-04
Fix to respect `RELEASE_MUTABLE_DIR` as erlsrv WorkDir

raymaung:
### Summary of changes

Fix **erlsrv** Windows service to respect `RELEASE_MUTABLE_DIR` as [erlsrv `WorkDir`](http://erlang.org/doc/man/erlsrv.html).

### DIFF
--- a/priv/templates/boot_win.eex
+++ b/priv/templates/boot_win.eex
@@ -19,7 +19,7 @@
 @if "%RELEASE_MUTABLE_DIR%"=="" (
   set mutable_dir=%release_root_dir%\var
 ) else (
-  set mutable_dir=%RELEASE_MUTABLE_DIR
+  set mutable_dir=%RELEASE_MUTABLE_DIR%
 )
 
 @if "%RELEASE_CONFIG_DIR%"=="" (
@@ -260,7 +260,7 @@
   %erlsrv% add %service_name% ^
             -c "%description%" ^
             %node_type% "%node_name%" ^
-            -w "%rootdir%\var" ^
+            -w "%mutable_dir%" ^
             -m "%start_erl%" ^
             -args "!service_args!" ^
             -debugtype new ^


JonRowe:
Some background on this, we're attempting to use distillery to run our elixir (phoenix) project as a windows service (for automated deployment reasons).

The recommended practise for this is to use the requisite ProgramData directory for the service for writing to disk (e.g. in our case we're trying to log to disk so `C:/ProgramData/MyPhoenixApp/log/production.log`) however the current distillery config is writing those files to the project directory which is not writable by default.

We can work around the location with environment variables but it needs to at least be the mutable directory and not the project root, we'd also recommend dropping var from the path as it means nothing to windows, but thats not a blocker for us if var is inside the program data folder.

Any advice on this would be thoroughly appreciated!

priv/templates/boot_win.eex
```diff
@@ -19,7 +19,7 @@
 @if "%RELEASE_MUTABLE_DIR%"=="" (
   set mutable_dir=%release_root_dir%\var
 ) else (
-  set mutable_dir=%RELEASE_MUTABLE_DIR
+  set mutable_dir=%RELEASE_MUTABLE_DIR%\var
```
bitwalker: I'm not sure we need to add `\var` here - the idea is that `RELEASE_MUTABLE_DIR` is already set to something like `path\to\var`, so we don't want to manipulate it. We do so in the "is empty" branch because `<root>\var` is the default when not set.

priv/templates/boot_win.eex
```diff
@@ -260,7 +260,7 @@
   %erlsrv% add %service_name% ^
             -c "%description%" ^
             %node_type% "%node_name%" ^
-            -w "%rootdir%\var" ^
+            -w "%mutable_dir%" ^
```
bitwalker: Good catch!

raymaung:
@bitwalker Updated to remove `\var` from `%mutable_dir%`.

bitwalker:
Thanks!

PR #327, opened 2017-09-13
Allow migrations by Repo

nardopessoa:
The "repo" name was fixed in sample code. I just put it dynamically using `Mix.Ecto` module.

### DIFF
--- a/docs/Running Migrations.md
+++ b/docs/Running Migrations.md
@@ -68,12 +68,18 @@ defmodule MyApp.ReleaseTasks do
   defp run_migrations_for(repo) do
     app = Keyword.get(repo.config, :otp_app)
     IO.puts "Running migrations for #{app}"
-    Ecto.Migrator.run(repo, migrations_path(app), :up, all: true)
+    Ecto.Migrator.run(repo, migrations_path(repo), :up, all: true)
   end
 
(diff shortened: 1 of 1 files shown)


michalmuskala:
The `Mix.Ecto` module is not public.

bitwalker:
Unfortunately, if `Mix.Ecto` is not public, I won't merge this, as it could cause problems if that module goes away or changes. Was there a specific problem with the old migrator example?

nardopessoa:
Well, the string "repo" used in example is a reference for Repo module name.
My Repo name is AppRepo and the folder for yours migrations is "priv/app_repo/migrations" (generated by Ecto.Gen.Migration).
When I run the example code, the folder "repo" never was found.
I fix this with this pull request codes... but... it's not ok... =(

bitwalker:
@nardopessoa If you submit a PR which passes both the app name and repo name to the affected functions, we should only need to change the use of `"repo"` to `Macro.underscore(repo)`.

nardopessoa:
@bitwalker, I fix the code using your sugestion: `Macro.underscore(repo)`.
I hope everything is right now.

bitwalker:
Thanks!

PR #330, opened 2017-09-20
Fix: release prod command

valentinogagliardi:
### Summary of changes

In "Building your release" section of Use with Phoenix the step 3 suggests running `mix release --env=prod` to build a production release. The command will generate a new release inside `_build/dev` which is not production.

The commit fixes the instructions to use `MIX_ENV=prod mix release` instead of `mix release --env=prod`

### Checklist

### DIFF
--- a/docs/Use With Phoenix.md
+++ b/docs/Use With Phoenix.md
@@ -70,7 +70,7 @@ steps:
 1. `MIX_ENV=prod mix phoenix.digest` To compress and tag your assets
     for proper caching. More detail can be found in the
     [Phoenix Mix Task Guide](http://www.phoenixframework.org/docs/mix-tasks#section--mix-phoenix-digest-)
-1. `MIX_ENV=prod mix release --env=prod` To actually generate a release for a
+1. `MIX_ENV=prod mix release` To actually generate a release for a
     production environment
 
 There are some optional flags available to you as well:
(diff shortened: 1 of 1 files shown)


aphillipo:
I do wonder if there is ever a reason to build a release in dev mode, it just seems doomed to not work with things like live reload, brunch or web pack, dev settings and probably a desire for mix phx.server to still work.  Getting dev mode to work with all of the above seems unlikely.

bitwalker:
With the default `rel/config.exs` file, setting `MIX_ENV=prod` will automatically imply `--env=prod`, but that's not true the other way around. The current behaviour is my preference. Phoenix is the most notable case of something which doesn't run well in `dev` (from a release), due to the code reloader primarily, but there are plenty of applications where that isn't an issue, so we can't make assumptions based on Phoenix alone.

@valentinogagliardi Can you remove the `--env=prod` from the instructions since MIX_ENV=prod is now explicit?

OvermindDL1:
On a side note, I'm really thinking a new MIX_ENV of `release` would be useful as a `release.exs` config file would solve some of phoenix's specific issues, if you do not mind building 2 releases (one for release, one for prod or so, I.E. one for running the server, the other for release tools so you can run migrations and such).  Just the start of an idea though...

bitwalker:
@OvermindDL1 Yeah that's definitely one recommended approach, and has the benefit of forcing you to put all the release specific tweaks to the configuration in one place. It does require modifying the default `mix.exs` though so that the `release` env gets the same project-level configuration (such as `build_embedded`, etc.) as the `prod` env.

valentinogagliardi:
@bitwalker done. Thank you!

bitwalker:
Thanks!

PR #331, opened 2017-09-20
Add boot opts

gfodor:
### Summary of changes

Adds a `boot_opts` config to allow additional keyword options to be sent to the boot loader assembler. We had a need to flip on the `no_dot_erlang` setting, so this was the patch I came up with. I am new to erlang and elixir so would like guidance on improving the PR if needed.

### Checklist

### DIFF
--- a/lib/mix/lib/releases/assembler.ex
+++ b/lib/mix/lib/releases/assembler.ex
@@ -711,7 +711,7 @@ defmodule Mix.Releases.Assembler do
   end
 
   # Generates .boot script
-  defp make_boot_script(%Release{profile: %Profile{output_dir: output_dir}} = release, rel_dir) do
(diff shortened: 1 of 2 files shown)


bitwalker:
Thanks for the PR! I'm generally in favor of this, but we may need a more subtle approach. We need to protect the `ERTS_LIB_DIR` variable, and I'm not sure (without reviewing the systools source) how it handles duplicate items in the option list. We probably also want to verify that the options provided are valid and return a proper error message if an invalid option is provided.

Other than that, we'll also need documentation on the new option, with a link to the Erlang docs for context; that way people know how to access and use it.

gfodor:
sorry for the delay on this, hopefully can get to this sometime soon!

bitwalker:
In master I've added support for `set no_dot_erlang: true` to allow making use of that option when building boot scripts; I prefer explicitly supporting options rather than allowing passing an arbitrary list - at least at this point.

PR #332, opened 2017-09-21
add RELEASE_ROOT_DIR to example so the migrate command always works

tverlaan:
This avoids searching for the correct environment variable and will make the migrate command just work(tm).


bitwalker:
Thanks!

PR #339, opened 2017-10-09
Fix Running Migrations doc example

joshprice:
Example was using `seed` rather than `migrate` as indicated by the shell script filename


bitwalker:
Thanks!

PR #344, opened 2017-10-12
Use consolidated protocols when installing windows service with erlsrv

thorsten-de:
When installing a release in Windows as a service, the `%consolidated_dir%` was not given to erlsrv.exe. Therefore, procotols are not consolidated and performance is hurt.

With this change, consolidated protocols are given to erlsrv as service_args during "erlsrv.exe add". They are used when the Windows service starts, and therefore performance is improved.


bitwalker:
Could you rebase this on master, there is a conflict. Thanks!

PR #350, opened 2017-10-17
Use @echo off instead of prefixing all commands with @ on Windows #349

### LINKED ISSUE #349: @call is not available on Windows 7
### Steps to reproduce

* Create any release in Windows 7 SP 1
* Try to run it

### Verbose Logs

You will get an error that `@call` is unknown and script will fail to run.

### Description of issue

- What are the expected results?

Script runs.

- What OS, Erlang/Elixir versions are you seeing this issue on?

Windows 7 SP1, Elixir 1.4.4, Erlang 19

mspanc:
### Summary of changes

The current syntax don't work on old Windows as `@call` is not available. That fix improves compatibility with old Windows (tested on Win 7 SP1 32-bit) and results in a cleaner code as `@echo off` is being called once.

Fixes #349.

### Checklist

### DIFF
--- a/priv/templates/boot_loader_win.eex
+++ b/priv/templates/boot_loader_win.eex
@@ -1,10 +1,11 @@
 :: Discover the current release directory from the directory
 :: of this script and the start_erl.data file
-@set script_dir=%~dp0
-@for %%A in ("%script_dir%\..") do @(
+@echo off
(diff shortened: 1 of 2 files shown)


bitwalker:
Thanks for this! I'll get it out in the next release :)

PR #356, opened 2017-10-21
add option for soft purge in appup make

redink:
## Summary of changes

Added one option for allow use `soft_purge` to generate `.appup` file and `relup` file.

### Checklist

just an example: https://github.com/redink/distillery_example

### DIFF
--- a/docs/Overlays.md
+++ b/docs/Overlays.md
@@ -31,7 +31,7 @@ Currently, the following overlay vars are provided out of the box by Distillery:
   - release_version: Version of the release being built
   - is_upgrade: Is this release an upgrade release
   - upgrade_from: The version of the release being upgraded from, nil if not an upgrade.
-  - soft_purge: Is this appup using soft purge when upgrade
+  - soft_purge: Should appups use soft purge during upgrade (see appup docs for more info)
   - dev_mode: Is this release being built in dev mode
   - include_erts: Is ERTS being included in the release
   - include_src: Is source code being included in the release
--- a/docs/Upgrades and Downgrades.md
+++ b/docs/Upgrades and Downgrades.md
@@ -89,4 +89,4 @@ more.
 
 To generate an upgrade release, you'll need to pass `--upgrade` to `mix release`. To generate an upgrade from an arbitrary
 version you've previously built, pass `--upfrom=<version>`. Distillery will look for the appup in the ebin of your current build.
-If you want soft purge during the upgrade/downgrade, you could use `--soft-purge`.
+If you want to use soft purge rather than the default brutal purge of old module versions during upgrade/downgrade, you can pass `--soft-purge` or set `soft_purge: true` in the release config file.
--- a/lib/distillery/tasks/release.ex
+++ b/lib/distillery/tasks/release.ex
@@ -19,9 +19,6 @@ defmodule Mix.Tasks.Release do
       # Build an upgrade release from a specific version
       mix release --upgrade --upfrom=0.1.0
 
-      # Build an upgrade release using soft purge
-      mix release --upgrade --soft-purge
-
       # Build a specific release
       mix release --name=myapp
 
--- a/lib/mix/lib/releases/appups.ex
+++ b/lib/mix/lib/releases/appups.ex
@@ -153,22 +153,32 @@ defmodule Mix.Releases.Appup do
   end
 
   # supervisor
-  defp generate_instruction_advanced(m, true, _is_special, _dep_mods, _), do: {:update, m, :supervisor}
+  defp generate_instruction_advanced(m, true, _is_special, _dep_mods, _) do
+    {:update, m, :supervisor}
+  end
   # special process (i.e. exports code_change/3 or system_code_change/4)
-  defp generate_instruction_advanced(m, _is_sup, true, [], false),        do: {:update, m, {:advanced, []}}
+  defp generate_instruction_advanced(m, _is_sup, true, [], false) do
+    {:update, m, {:advanced, []}}
+  end
   defp generate_instruction_advanced(m, _is_sup, true, [], true) do
     {:update, m, {:advanced, []}, :soft_purge, :soft_purge, []}
   end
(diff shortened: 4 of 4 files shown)


docs/Overlays.md
```diff
@@ -31,6 +31,7 @@ Currently, the following overlay vars are provided out of the box by Distillery:
   - release_version: Version of the release being built
   - is_upgrade: Is this release an upgrade release
   - upgrade_from: The version of the release being upgraded from, nil if not an upgrade.
+  - soft_purge: Is this appup using soft purge when upgrade
```
bitwalker: Let's change this a bit to "Should appups use soft purge during upgrade (see appup docs for more info)"

docs/Upgrades and Downgrades.md
```diff
@@ -89,3 +89,4 @@ more.
 
 To generate an upgrade release, you'll need to pass `--upgrade` to `mix release`. To generate an upgrade from an arbitrary
 version you've previously built, pass `--upfrom=<version>`. Distillery will look for the appup in the ebin of your current build.
+If you want soft purge during the upgrade/downgrade, you could use `--soft-purge`.
```
bitwalker: Can you expand the docs to here to explain the difference between soft purge and brutal purge, and why you would choose soft purges? It will certainly be the first thing people will want to know when they see the option :).

I would change the text here to read "If you want to use soft purges rather than the default brutal purge of old module versions during upgrades/downgrades, you can pass `--soft-purge` or set `soft_purge: true` in the release config file."

lib/distillery/tasks/release.ex
```diff
@@ -19,6 +19,9 @@ defmodule Mix.Tasks.Release do
       # Build an upgrade release from a specific version
       mix release --upgrade --upfrom=0.1.0
 
+      # Build an upgrade release using soft purge
```
bitwalker: I would rather not document this option here, as it is really an advanced option that one should consult the documentation on using first. This keeps the output of `mix help release` more useful (generally the folks running help are using Distillery for the first time, or when they are still new to releases).

lib/mix/lib/releases/appups.ex
```diff
@@ -149,13 +153,25 @@ defmodule Mix.Releases.Appup do
   end
 
   # supervisor
-  defp generate_instruction_advanced(m, true, _is_special, _dep_mods), do: {:update, m, :supervisor}
+  defp generate_instruction_advanced(m, true, _is_special, _dep_mods, _), do: {:update, m, :supervisor}
   # special process (i.e. exports code_change/3 or system_code_change/4)
-  defp generate_instruction_advanced(m, _is_sup, true, []),       do: {:update, m, {:advanced, []}}
-  defp generate_instruction_advanced(m, _is_sup, true, dep_mods), do: {:update, m, {:advanced, []}, dep_mods}
+  defp generate_instruction_advanced(m, _is_sup, true, [], false),        do: {:update, m, {:advanced, []}}
```
bitwalker: Can you change the formatting to use two lines for any of these on a single line? Example:

```elixir
defp generate_instruction_advanced(m, _is_sup, true, [], false), 
  do: {:update, m, {:advanced, []}}
```

redink:
Thanks for review !

redink:
I am not sure the reason for timeout, need help. 😞

redink:
ping ...

redink:
@bitwalker ping

redink:
I think we can control the purging behavior for specific application by the configuration:

```elixir
config :distillery,
  app1: :soft,
```

so, we can use `soft_purge` when make appup, but I think it make little sense. If some application(modules) using soft purge and others using brutal purge, the finally result will be affected by soft purge. Because if some process using old version code, the scheduler will wait the process suspend. Brutal purge just only kill the processes which using old version code and can not affect the final result of hot upgrade.

And it is too expensive to control every modules purging behavior. If someone has such demand, maybe he/she should modify the relup file manually.

bitwalker:
@redink The issue is that we're getting into territory where people (like yourself) want more control over how the appup is generated because the defaults aren't working for them. On the one hand, I'm tempted to just say that if the defaults don't work, you need to modify the appup by hand, keeping Distillery simple. If we are going to provide knobs for controlling the appup generator, then we need to implement them in such a way that it is extendable so that other options can be specified without requiring changes in a bunch of places, and we need to ensure that if we provide an option, that it's at the right level of abstraction.

In the case of soft vs brutal purges, you can certainly mix them without issues, and while I agree that typically it would make more sense to change the value globally, that's certainly not the case all the time. So if we're going to provide the option, the right level of abstraction is per-application. But we can get a cross between the two, where one can set the default, and then override per-application. Such configuration could look something like this:

```elixir
# rel/config.exs

release :foo do
  set appup_options: [
    default_purge_type: :brutal_purge,
    purge_overrides: [app1: :soft_purge]
  ]
end
```

I'm open to the above, but what I think I'd really prefer more than anything is a means to plug in transformation passes, so your config would look something like this:

```elixir
# rel/config.exs

release :foo do
  set appup_transforms: [
    {Foo.Appups.PurgeTransform, [default: :soft_purge, overrides: [app1: :brutal_purge]]}
  ]
end

# foo/lib/appups/purge_transform.ex

defmodule Foo.Appups.PurgeTransform do
  use Distillery.Appups.Transform

   def upfrom(v1, v2, {:update, m, {:advanced, args}, dep_mods}) do
     {:update, m, {:advanced, args}, :soft_purge, :soft_purge, deps_mods}
   end
   def upfrom(_v1, _v2, instruction), do: instruction
end
```

The above is a more maintainable approach because it automates the transformation, and can be extended arbitrarily to do whatever you want, including adding additional instructions to the appup. Ultimately that's a better approach, as it provides a means for libraries to provide specialized transformations for themselves (or for other libraries) - for example, a Phoenix transformation that handles hot upgrades between major releases.

So ultimately I think I'd rather back up a step and rethink how we want to expose customizing the appup generators behaviour, and go that direction, rather than add more ad-hoc options to the config.

bitwalker:
Closing this in favor of #403

PR #360, opened 2017-10-26
Allow to ignore mutable directory creation by setting special envirom…

liveforeverx:
…ents

### Summary of changes

Example, based on `app.sh` you want to provide some functionallity, which can be executed under another users, as release was started.

For example, you start release with User A, than User B uses escript, but User B can't use escript, because he can't access `var/` folder. Even he can access, than he can override it, if escript was used before starting of release, than release want started, because it can't override files from User B.

To allow using some commands under another user, as the release running.

### Checklist

So, I would like to implement better general solution, but I need advice about a way.

### DIFF
--- a/priv/libexec/config.sh
+++ b/priv/libexec/config.sh
@@ -30,11 +30,15 @@ configure_release() {
         export SRC_VMARGS_PATH="$VMARGS_PATH"
     fi
     if [ "$SRC_VMARGS_PATH" != "$RELEASE_MUTABLE_DIR/vm.args" ]; then
-        echo "#### Generated - edit/create $RELEASE_CONFIG_DIR/vm.args instead." \
-            >  "$RELEASE_MUTABLE_DIR/vm.args"
-        cat  "$SRC_VMARGS_PATH"                              \
-            >> "$RELEASE_MUTABLE_DIR/vm.args"
-        export DEST_VMARGS_PATH="$RELEASE_MUTABLE_DIR"/vm.args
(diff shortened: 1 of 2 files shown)


bitwalker:
I think we should probably make it so that if files under `/var` exist and are readable but not writable, then we skip trying to generate fresh copies of those files.

That said, I'm not sure I understand the threat model here where someone can be privileged enough to read everything in the release directory, but not enough where they can be given read/write access to `/var`. As long as the original source files are marked read-only for that user, no matter what changes they make in `/var`, the files will just be regenerated anyway.

liveforeverx:
Hi!

Example: there is an `app` user. Application is started with app user. There exists administration `CLI`, which allows to run some commands and admin logged in as root. Of course he can rewrite config files as `root`, but afterwards `app` willn't be possible to start, because admin execution of any command on cli via `root` destroyed the configuration rights. Means. Simple get any information via `CLI` via `root` makes it non startable for `app` user for application themself.

So, for this case, we want to allow admin (which is logged via `root` or any other user) to use administrative commands (check application/cluster state, put some information to app) without changing of rights for config `var` - folder.

bitwalker:
I see, I think I'm going to modify this a bit to be slightly more general - basically provide a `RELEASE_READ_ONLY` var which will skip anything which requires writing/modifying what's on disk, for the purpose of running commands, etc in the context you described. If anything is required and missing, then an error will be raised when `RELEASE_READ_ONLY=true`. I'm not sure if there are many more places other than what is touched in this PR, but I want to be sure of that.

Issue #361, opened 2017-10-30
Code in `Running migrations, etc` doc is broken.


chenxsan:
```
$ mix hex.info
Hex:    0.17.1
Elixir: 1.5.1
OTP:    20.1

Built with: Elixir 1.5.1 and OTP 18.3
```

I'm doing migration according to https://github.com/bitwalker/distillery/blob/8ec93745f06284ff70bf4c14187af11e56f461ef/docs/Running%20Migrations.md, and I'm afraid the example is totally broken.

Here's what I have in my `ReleaseTasks` module:

```
defmodule Myflow.ReleaseTasks do
  
  @start_apps [
    :crypto,
    :ssl,
    :mariaex,
    :ecto
  ]

  def myapp, do: Application.get_application(__MODULE__)

  def repos, do: Application.get_env(myapp(), :ecto_repos, [])

  def seed do
    me = myapp()

    IO.puts "Loading #{me}.."
    # Load the code for myapp, but don't start it
    :ok = Application.load(me)

    IO.puts "Starting dependencies.."
    # Start apps necessary for executing migrations
    Enum.each(@start_apps, &Application.ensure_all_started/1)

    # Start the Repo(s) for myapp
    IO.puts "Starting repos.."
    Enum.each(repos(), &(&1.start_link(pool_size: 1)))

    # Run migrations
    migrate()

    # Run seed script
    Enum.each(repos(), &run_seeds_for/1)

    # Signal shutdown
    IO.puts "Success!"
    :init.stop()
  end

  def migrate, do: Enum.each(repos(), &run_migrations_for/1)

  def priv_dir(app), do: "#{:code.priv_dir(app)}"

  defp run_migrations_for(repo) do
    app = Keyword.get(repo.config, :otp_app)
    IO.puts "Running migrations for #{app}"
    Ecto.Migrator.run(repo, migrations_path(repo), :up, all: true)
  end

  def run_seeds_for(repo) do
    # Run the seed script if it exists
    seed_script = seeds_path(repo)
    if File.exists?(seed_script) do
      IO.puts "Running seed script.."
      Code.eval_file(seed_script)
    end
  end

  def migrations_path(repo), do: priv_path_for(repo, "migrations")

  def seeds_path(repo), do: priv_path_for(repo, "seeds.exs")

  def priv_path_for(repo, filename) do
    app = Keyword.get(repo.config, :otp_app)
    repo_underscore = repo |> Module.split |> List.last |> Macro.underscore
[...]

PR #362, opened 2017-10-30
fix broken example

chenxsan:
Detail here https://github.com/bitwalker/distillery/issues/361, I've tested it on macOS & Ubuntu 16.04.


bitwalker:
Thanks!

PR #364, opened 2017-11-03
Don't change application order by hand, let allow `systools` does it

surik:
### Summary of changes

This change fixes `write_relfile` to generate `.rel` as simple as it possible. `systools` has the ability to make topological sorting for the applications by itself.

### Checklist


bitwalker:
Thanks! The other tests are broken right now, but I'll assume you did your due diligence for now. I believe there was a reason why this code was there previously, but after review I don't think it's necessary any longer.

PR #366, opened 2017-11-08
fixes typos

jaimeiniesta:
- it's -> its
- there *is* no significant differences -> there *are* no significant differences
- an upgrades -> an upgrade


bitwalker:
Thanks!

PR #373, opened 2017-11-28
Support REPLACE_OS_VARS on a custom vm.args.

jesseshieh:
### Summary of changes

Hi,

I'm trying to use a custom vm.args file using `VMARGS_PATH`, but I also want to use `REPLACE_OS_VARS=true` to replace variables inside my custom vm.args. I couldn't get it to work, I think, because `_configure_node` loads the node name from `VMARGS_PATH` instead of `DEST_VMARGS_PATH`. `DEST_VMARGS_PATH` has the replaced variables, but `VMARGS_PATH` is the original.

Let me know if I'm going about this the wrong way and I'll fix it. I haven't really tested this change thoroughly so it is really more of a proposal/question than an actual pull request. If you like the change, I'll test it properly.

The existing tests (`mix test`) pass, but I didn't see any tests for `config.sh` aside from integration tests. I could try and add an integration test that uses `VMARGS_PATH`?

Thanks!

### Checklist

### DIFF
--- a/priv/libexec/config.sh
+++ b/priv/libexec/config.sh
@@ -39,7 +39,7 @@ configure_release() {
     if [ ! -z "$REPLACE_OS_VARS" ]; then
         _replace_os_vars "$DEST_VMARGS_PATH"
     fi
-    export VMARGS_PATH="${VMARGS_PATH:-$DEST_VMARGS_PATH}"
+    export VMARGS_PATH="${DEST_VMARGS_PATH:-$VMARGS_PATH}"
 
     # Set SYS_CONFIG_PATH, the path to the sys.config file to use
     # Use $RELEASE_CONFIG_DIR/sys.config if exists, otherwise releases/VSN/sys.config


jesseshieh:
Hi @bitwalker, have you had a chance to look at this yet? I went through the checklist and I don't think this changes any typespecs, doesn't require any documentation changes, and the commits are squashed. This doesn't add a new test, but I couldn't really find any existing tests around this part of the code.

I also discovered another issue that this PR might fix.

If you use `RELEASE_CONFIG_DIR` to specify the location of a `vm.args` file, but you also want to use `REPLACE_OS_VARS=true`, all works great. If you then try to hot upgrade, though, the variable substitution no longer happens because `install.sh` sets the `VMARGS_PATH` before calling `configure_release`. `configure_release` then ignores the `DEST_VMARGS_PATH` similar to the original issue in this PR.

If you like, I can go into a bit more detail about what exactly is happening in the code line-by-line. It's kind of confusing what `VMARGS_PATH`, `SRC_VMARGS_PATH`, `DEST_VMARGS_PATH`, etc are and how they interact with each other.

bitwalker:
So sorry for the delay - we want to use `VMARGS_PATH` in the locations you've changed here, because `DEST_VMARGS_PATH` is something only the configure/install stuff should be aware of. It sounds like the correct fix is to address the interaction in the install command which is not setting `VMARGS_PATH` after writing out `DEST_VMARGS_PATH`. The thing to keep in mind while reading this code is that it is re-entrant when run as a daemon (i.e. via `bin/myapp start`), because `run_erl` uses the boot script to boot the actual release, but inherits the environment, so environment variables like `VMARGS_PATH` are set when `run_erl` executes. This makes it tricky to consider all of the interactions, but I think in this case we're just forgetting to set `VMARGS_PATH` the last time.

jesseshieh:
@bitwalker thanks for the thoughtful response! I changed this PR with an alternative fix. What do you think about this approach instead? `mix test` passes, but I haven't tested it thoroughly yet.

bitwalker:
@jesseshieh Can I ask how you validated the fix? The key thing is that we have to make sure that the path to the args file is correct during an upgrade (more specifically that the vm.args file in `var` matches the newest version from the upgraded release with all of the vars replaced). I think we can merge this, but I want to get a sense of how it was tested first (the current test suite doesn't validate that the updated vm.args is correct).

jesseshieh:
@bitwalker at the risk of too much information, here's what I did

1. set my local `mix.exs`'s distillery `path` to my local version with the change above
2. run `MIX_ENV=prod mix release --env=prod`
3. copy the release tarball to `/tmp/delete-me`
4. created a `/tmp/delete-me/custom_vm.args` with the contents
    ```
    -name custom-${MY_NODE_NAME}
    -setcookie custom-${MY_COOKIE}
    ```
3. start the app with 
    ```
    VMARGS_PATH=custom_vm.args DATABASE_URL="postgresql://user:pass@localhost:5432/foo" MY_HOSTNAME=example.com MY_COOKIE=secret REPLACE_OS_VARS=true MY_NODE_NAME=foo@127.0.0.1 PORT=4000 bin/gigalixir_getting_started console
    ```
4. I check that the node name is right and the cookie is right. In this case, they are 
    ```
    iex(custom-foo@127.0.0.1)5> Node.self      
    :"custom-foo@127.0.0.1"
    iex(custom-foo@127.0.0.1)6> Node.get_cookie
    :"custom-secret"
    ```

Then to test upgrades
1. bump version number in `mix.exs` to 0.0.2
2. run `MIX_ENV=prod mix release --env=prod --upgrade`
3. upgrade by running
    ````
    cd /tmp/delete-me
    cp /home/js/Development/gigalixir-getting-started/_build/prod/rel/gigalixir_getting_started/releases/0.0.2/gigalixir_getting_started.tar.gz releases/0.0.2/
    VMARGS_PATH=custom_vm.args DATABASE_URL="postgresql://user:pass@localhost:5432/foo" MY_HOSTNAME=example.com MY_COOKIE=secret REPLACE_OS_VARS=true MY_NODE_NAME=foo@127.0.0.1 PORT=4000 bin/gigalixir_getting_started upgrade 0.0.2
    ```
4. verify 0.0.2 is running fine. node name and cookie are unchanged.

I briefly tested a few other scenarios like
1. using environment `set vm_arg` directive
2. not specify any `vm.args` file at all anywhere which I think uses a default one

I didn't test things like
1. `RELEASE_CONFIG_DIR`
2. any `sys.config` stuff

jesseshieh:
Oh sorry, I checked `var/vm.args` and it looks good after an upgrade.
```
$ cat var/vm.args 
#### Generated - edit/create /tmp/delete-me/vm.args instead.
-name custom-foo@127.0.0.1
-setcookie custom-secret
```

But note that 0.0.1 and 0.0.2 both had `VMARGS_PATH` set to the same file. I'm not sure what would happen if I switched it to a different file during an upgrade.

bitwalker:
No level of information is too much ;)

I think this is probably safe, you've tested it more or less how I would have, except I probably would've changed the values in the args file between versions, but I can do that locally. Thanks for looking into this fix and doing the work to test it!

PR #376, opened 2017-12-06
Fixes Windows batch script

cgudrian:
commit cf22cd1da5b44f26221b5007919a905187d942d7
Author: Christian Gudrian <christian.gudrian@aucos.de>
Date:   Wed Dec 6 09:09:45 2017 +0100

    Fixes if statements

    "exists" was used instead of "exist".

commit e34540981f8deaee9134169f5a612e6628f58897
Author: Christian Gudrian <christian.gudrian@aucos.de>
Date:   Wed Dec 6 09:09:27 2017 +0100

    Fixes variable reference

    %root_dir% has been renamed to %release_root_dir%.

commit 322d73da2e56cae3f82badf504aa6ce0234772f2
Author: Christian Gudrian <christian.gudrian@aucos.de>
Date:   Wed Dec 6 09:08:57 2017 +0100

    Changes EOL style of the whole file

    Windows batch files need CRLF. Otherwise cmd.exe complains about
    unresolved labels.

    https://serverfault.com/questions/429594/is-it-safe-to-write-batch-files-with-unix-line-endings

commit 9f138224a692cf3e067add3233acfae9035c5dd1
Author: Christian Gudrian <christian.gudrian@aucos.de>
Date:   Tue Dec 5 14:40:38 2017 +0100

    Makes local pid variable available globally

commit d41c0fbc17590b86e0194942cbd83ca3d42db08e
Author: Christian Gudrian <christian.gudrian@aucos.de>
Date:   Tue Dec 5 14:40:04 2017 +0100

    Adds "usebackq" option to for loop

commit 062eb18b68aff4b9754b543ff0cc8c40f5d694dc
Author: Christian Gudrian <christian.gudrian@aucos.de>
Date:   Tue Dec 5 14:39:42 2017 +0100

    Fixes quoting

    A single quotation mark was used instead of a double one.

commit 60e98925c80b8c2b853f9fe099a9c1c417a19f39
Author: Christian Gudrian <christian.gudrian@aucos.de>
Date:   Tue Dec 5 14:39:14 2017 +0100

    Make line endings consistent

    Some lines were using CRLF.

### DIFF
--- a/priv/templates/boot_win.eex
+++ b/priv/templates/boot_win.eex
@@ -167,7 +167,7 @@ if exist %possible_sys% (
 )
 set sys_config_target=%config_dir%\sys.config
 :: Replace environment variables
-powershell -command "$content = get-content %sys_config_src%; [regex]::matches($content, '\${[\w\d_-]+}') | foreach { $content = $content.Replace($_.value, [System.Environment]::GetEnvironmentVariable($_.value)) }; out-file -filepath %sys_config_target% -inputobject $content -encoding ascii" 
+powershell -command "$content = get-content %sys_config_src%; [regex]::matches($content, '\${[\w\d_-]+}') | foreach { $content = $content.Replace($_.value, [System.Environment]::GetEnvironmentVariable($_.value)) }; out-file -filepath %sys_config_target% -inputobject $content -encoding ascii"
 set sys_config=%sys_config_target%
 goto :eof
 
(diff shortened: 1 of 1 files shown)


sparta-developers:
One issue that we just uncovered (which does not seem related to this pull request, but which you might be able to quickly diagnose, since you seem much more familiar with these templates) is that when installing into a directory where the name includes whitespace characters, we are unable to start the application.

For example, with an application named `distillery_app`, if we untar the release into `C:\Program Files\distillery_app` and run `"C:\Program Files\distillery_app\bin\distillery_app" install`, we receive the following error message:

```
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
```

We suspect this is because of quoting (or lack thereof) somewhere in the template, but in the short term are going to install the application in a different directory.

bitwalker:
I hate to do this, but can you submit this PR "unsquashed"? The diff is completely unusable in this case because effectively the entire file has changed. In order for me to properly review it, I need to see the commits where things other than line endings changed to do a proper review. I generally make the decision to squash or not myself when merging, so you don't need to do that yourself. Feel free to squash commits which all change line endings though, as long as those commits don't also include other changes. That kind of thing is fine since it generally doesn't impact readability.

I'll leave this open for now, in case you don't have the unsquashed changes any longer, but in that case I'll need you to point out the changes you made and hope nothing is missed.

cgudrian:
Closing in favor of #378.

PR #377, opened 2017-12-06
Fix quotes in boot_win template

sparta-developers:
### Summary of changes

The templates that were generated for Windows had incorrect quotes.

See https://stackoverflow.com/questions/47624122/elixir-application-quits-immediately-or-doesnt-receive-input-when-running-dis/47663935 and https://www.bountysource.com/issues/50408118-starting-in-foreground-mode-on-windows-fails-to-get-pid-and-returns-errorlevel-1

### Checklist


sparta-developers:
Déjà vu! It looks like #376 solves our problem, and much more. We'll close this pull request in favor of that one.

PR #378, opened 2017-12-08
Fixes Windows batch script

cgudrian:
These commits fix issues in the batch template for Windows.

This is the unsquashed (and slightly revised) version of #376.


bitwalker:
Thanks for the fixes! It is much appreciated :)

PR #380, opened 2017-12-14
Boot win fixes

TuongNM:
### Summary of changes

I replaced the sleep command with timeout since the sleep command couldn't be found on my Windows 10 Pro (apparently it's only available if you install the Resource Kits). The timeout command is available since Windows Vista.

I also redirected the tasklist output to nul so that it doesn't fill the command prompt.

### Checklist


bitwalker:
Thanks!

PR #382, opened 2017-12-22
Adds commands to init task

aditya7iyengar:
### Summary of changes

We are working on an Elixir Deploy automation library, [Akd](https://github.com/annkissam/akd/pull/2/files). As part of the build process it calls `mix release.init`. We needed to add some custom commands to `rel/config.exs`. Instead of modifying the generated config file, we were hoping to add those commands to the mix task itself. This would allow `Akd` to dynamically create configs.

### Checklist

### DIFF
--- a/lib/distillery/tasks/init.ex
+++ b/lib/distillery/tasks/init.ex
@@ -26,6 +26,9 @@ defmodule Mix.Tasks.Release.Init do
       # invalid characters replaced or stripped out.
       mix release.init --name foobar
 
+      # Initializes releases with custom commands.
+      mix release.init --command cmd:path/to/file --command anothercmd:path/to/anotherfile
+
       # Use a custom template for generating the release config.
       mix release.init --template path/to/template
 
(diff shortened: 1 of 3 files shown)


bitwalker:
Seems to me that if you need to autogenerate some custom commands, a plugin is a better path. For one, if someone wants to remove the library, they won't have to also clean up their config file, and secondly, it allows you to modify the commands as needed, rather than relying on the user to update their config file. Can you explain why this approach is necessary?

aditya7iyengar:
Plugin seems like a great alternative. Although, one of the things we are trying to accomplish is to use commands which can take other commands as arguments and invoke them. For example, one of our custom commands is `env` which can invoke other commands with a list of environment variables:
`$ bin/app env start`

Doesn't look like this can be accomplished by Plugins. Let me know if I'm wrong though.. :smile:

bitwalker:
Sorry, what I meant by plugins are Distillery plugins, that execute as part of the release build pipeline. Using a plugin, `akd` would modify the configuration programmatically in the plugin's `before_assembly` callback, by injecting the custom commands in the Config structure used by Distillery. This would avoid the need to generate a config file, would allow adding `akd` to an existing project without any manual steps, and would allow you to make changes between versions without the user having to modify the config file manually. Make sense?

bitwalker:
I'm closing this as I believe this problem can be solved better using the plugin system. I recently created a branch which adds a plugin like this to Distillery to bundle a new `edump` command, called `edump_plugin`, give that a look if you are interested in how to accomplish your goal using plugins.

PR #390, opened 2018-01-16
Update extended-grep call to use egrep instead

larryweya:
The -E switch is not a valid switch on smartos grep

### Summary of changes

### Checklist


bitwalker:
Thanks!

PR #392, opened 2018-01-25
modify boot_win.eex to prevent exit statements to close the shell calling bat file

oboudry:
### Summary of changes

Replacement of all `exit` statements with `exit /b` in the boot_win.eex script that is used to create windows .bat files for a release. The current `exit` statements cause the calling shell to close, which is very annoying and prevents user from reading error messages outputed by the .bat file. Adding the /b flag will cause the batch script to stop but not the MS-DOS shell that called the script.


bitwalker:
Good call, thanks!

PR #394, opened 2018-01-30
Reformat systemd docs to render dorrectly 

michalmuskala:
* Change the format of systemd docs to render correctly. Right now they look very bad:

screen shot 2018-01-30 at 12 53 19

* Also update ex_doc.


bitwalker:
Thanks!

PR #397, opened 2018-02-13
Sanitize release name for umbrella projects

joaquimadraz:
### The problem
I was checking adding Distillery to my umbrella project and the repository has the domain in the name, eg: `distillery.io`. 
From what I could grasp, for umbrella projects Distillery uses the folder name as release name.
When running `mix release.init`, for the given example `distillery.io` the configuration is invalid and does not compile:
```elixir
  release :distillery.io do
    set version: "0.1.0"
    set applications: [
      :runtime_tools,
      distillery: :permanent
  ]
```
### Summary of changes
Sanitize the folder name allowing only letters, numbers and underscores, everything else is replaced by underscore.
I've extracted the sanitization logic to the `Utils` module.
As for the test, I don't really like the setup but I'll wait for your feedback on that 👍 
### Checklist


lib/mix/lib/releases/utils.ex
```diff
@@ -438,4 +438,25 @@ defmodule Mix.Releases.Utils do
...
+      "project_name"
+
+      iex> #{__MODULE__}.to_underscore("project name")
+      "project_name"
+
+      iex> #{__MODULE__}.to_underscore("project-name")
+      "project_name"
+  """
+  @spec to_underscore(String.t) :: String.t
+  def to_underscore(text) do
```
bitwalker: Might make more sense to name this `sanitize_name`, since the name doesn't really reflect what it does.

bitwalker:
I think a better test here would be to create a project on the fly, build a release in it, and make sure it starts. This doubles as a test for the process of going from new project to release. I don't think it's necessary to check it in (it's a lot of bloat just to check that it reads the config properly, there is probably a much easier method for doing that much).

bitwalker:
It occurs to me that perhaps a better approach here would be to just quote the atom, rather than sanitize it. That said, we probably do want to perform some degree of sanitization, since the name is used in a variety of contexts, and it is possible something else might get screwed up with really weird names. So I think this is probably a good approach 👍

joaquimadraz:
Just noticed that `distillery.io` is actually an invalid name for an Elixir application:

```
❯ mix new distillery.io --umbrella
** (Mix) Application name must start with a letter and have only lowercase letters, numbers and underscore, got: "distillery.io". (...)
```

I found the problem because I changed my project's name on Github and when cloning it the folder was the domain name. It looks like this is an edge case. It's still worth it to sanitize the name though.

So for the test, I'll create a project and rename the folder before running distillery. Sounds good?

bitwalker:
Yeah I know Mix won't allow it, but like you said, the edge case is always possible, so it's worth it to sanitize. As for the test, sounds good!

joaquimadraz:
I've just pushed the changes based on your feedback. 
To add `distillery` as dependency I'm doing some string replacements on `mix.exs` which feels a bit clunky but I didn't find any other way of doing it.
How do you feel we should go about it?

joaquimadraz:
Hey @bitwalker, do you have any feedback on this?

bitwalker:
I've taken a different approach on this. Rather than guessing the app names based on directory, the init task now actually parses/evaluates the AST of the `mix.exs` file in each app, and gets the application name from the project definition. This way we can ensure we don't have to replicate Mix semantics for names, we just use whatever was defined.

joaquimadraz:
Yeah, sounds like a better approach 👍

PR #405, opened 2018-03-25
WiP: Documentation fixes

PhillippOhlandt:
As promised in Slack, I gonna go through the whole documentation. Found some typos on my way and thought it would be good to create a WiP PR where I fix those as I move along. 

Gonna squash them in the end.

### DIFF
--- a/docs/introduction/terminology.md
+++ b/docs/introduction/terminology.md
@@ -90,7 +90,7 @@ For lack of a better name, it is how we currently refer to the script executed w
 #### Boot Script
 
 The script containing instructions for the VM on how to boot. The source form has the `.script` extension, and the
-"compiled" or binary form of the script has the `.boot` extension. You can get the source from from the binary form
+"compiled" or binary form of the script has the `.boot` extension. You can get the source form from the binary form
 by piping it through `:erlang.binary_to_term/1`.
 
 #### Boot Hook
--- a/docs/guides/phoenix_walkthrough.md
+++ b/docs/guides/phoenix_walkthrough.md
@@ -39,7 +39,7 @@ Next we will add Distillery to the deps function of our `mix.exs` file.
 First let's modify the Phoenix `config/prod.exs` file. Change this section of text:
 
 ```
-config :phx_distillery, PhxDistilleryWeb.Endpoint,
+config :phoenix_distillery, PhoenixDistilleryWeb.Endpoint,
   load_from_system_env: true,
   url: [host: "example.com", port: 80],
   cache_static_manifest: "priv/static/cache_manifest.json"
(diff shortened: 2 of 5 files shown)


PhillippOhlandt:
In `introduction/terminology` there is the following:

```
#### Appup

A file containing Erlang terms which describes with high-level instructions how to upgrade and downgrade
between the current release and one or more older releases.

#### Relup

A file containing Erlang terms which describes with low-level instructions how to upgrade and downgrade
between the current release and one or more older releases.
```

As a reader, this is a bit confusing because it's the same text for two things. Maybe this can be improved.

docs/introduction/terminology.md
```diff
@@ -90,7 +90,7 @@ For lack of a better name, it is how we currently refer to the script executed w
 #### Boot Script
 
 The script containing instructions for the VM on how to boot. The source form has the `.script` extension, and the
-"compiled" or binary form of the script has the `.boot` extension. You can get the source from from the binary form
+"compiled" or binary form of the script has the `.boot` extension. You can get the source form from the binary form
```
bitwalker: Why the change from `from` to `form`?

PhillippOhlandt: We could also remove the first "from/form" and it would still make sense. "You can get the source from the binary form"

bitwalker: Lets phrase it like "You can get back the script source from the binary..."

bitwalker:
@PhillippOhlandt Well the explanation is _slightly_ different, do you mean it is still confusing even though they refer to two different levels of abstraction?

bitwalker:
We may also want to address this by reordering the documentation somewhat, or moving terminology into individual documents into some kind of lead-in section and removing the big glossary doc.

docs/introduction/terminology.md
PhillippOhlandt: Sounds good. Gonna include that in my next round of changes.

PhillippOhlandt:
> Well the explanation is slightly different, do you mean it is still confusing even though they refer to two different levels of abstraction?

Oh boy, just noticed the high-level/low-level wording. Completely missed it. So I think it might be fine. Not sure if others are as blind as me, tho.

> We may also want to address this by reordering the documentation somewhat, or moving terminology into individual documents into some kind of lead-in section and removing the big glossary doc.

We could. I am not done with reading everything so I can't really say anything about ordering yet.

PhillippOhlandt:
I just read the "Appups" chapter. However, I wasn't able to find the `.appup` file for the project I got from the Phoenix Walkthrough. All I have there is the `.app` file.

bitwalker:
@PhillippOhlandt The .appup will be generated into the `_build` directory, in the `ebin` directory for the application being upgraded. There are changes coming to this side of things though, so perhaps we can hold off on fixing that until 2.0 is released (soon)

bitwalker:
Sounds good, thanks for all the help!

PR #415, opened 2018-05-17
Doc: improved code example for running migrations and seeds

andreapavoni:
### Summary of changes

As already mentioned in a [tweet](https://twitter.com/apeacox/status/995949484461973504), I'm proposing an improved code example to run migrations and seeds. The first time I approached the [original code example](https://hexdocs.pm/distillery/running-migrations.html), I had an hard time to understand it correclty. In particular, I didn't noted that the main logic to start the processes was running under the `seed/0` function, while `migrate/0` was pretty unusable alone. Moreover, from a code-style point of view, the only _public_ functions should have been `seed/0` and `migrate/0`.

So, I've refactored the example for my own project, and now I'd would be glad to share this snippet with other users, hoping they'll find it more useful/usable than the actual one.


bitwalker:
Thanks!

PR #417, opened 2018-05-27
Make sure distillery app is started before the main release app, when needed

studzien:
### Summary of changes

It seems that starting from 25eb0b80f92c0fcbafa9c7293069065dccdfd219 the main release application is not started, if `applications` value is inferred automatically from the deps in mix.exs (see the test release, released from `standard_app/extra_apps.mix.exs`), because:

1. the main app ends up having `distillery` in application dependencies, but the generated boot script doesn't try to start it before the main app;
2. that's because the start type from initially set `{:distillery, :load}` is not changed, when processing the deps of the main app.

What was kind of surprising to me is that the boot process doesn't fail, we just end up with Erlang VM running, but without the main app.

The proposed fix is kind of the simplest thing that seems to work in case of explicitly passed `applications` and when it's inferred.

The other problem I can see (this kind of applies to adding the call to distillery somewhere in the boot process) is that we never can add distillery with the `runtime: false` option, as this will always remove it from applications included in the release.

### Checklist

### DIFF
--- a/lib/mix/lib/releases/models/release.ex
+++ b/lib/mix/lib/releases/models/release.ex
@@ -12,9 +12,7 @@ defmodule Mix.Releases.Release do
               # included so the elixir shell works
               :iex,
               # required for upgrades
-              :sasl,
-              {:mix, :load},
-              {:distillery, :load}
+              :sasl
               # can also use `app_name: type`, as in `some_dep: load`,
               # to only load the application, not start it
             ],
--- a/lib/mix/lib/releases/utils.ex
+++ b/lib/mix/lib/releases/utils.ex
@@ -346,6 +346,11 @@ defmodule Mix.Releases.Utils do
           end
       end)
 
+    base_apps =
+      base_apps
+      |> add_app_if_doesnt_exist(:mix, :load)
+      |> add_app_if_doesnt_exist(:distillery, :load)
+
     # Correct any ERTS libs which should be pulled from the correct
     # ERTS directory, not from the current environment.
     apps =
@@ -383,6 +388,7 @@ defmodule Mix.Releases.Utils do
               end
           end)
       end
+
 
     case apps do
       {:error, _} = err ->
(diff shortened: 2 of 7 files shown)


studzien:
I've spent some time investigating failing Travis job (I was able to reproduce it on my machine, given enough test runs) and it seems that it was due to the race condition between calling `start` in `rpc` in the test.

The outcome was that sometimes the `var/sys.config` file in the `standard_app` had the following, non-parsable, format:
```
%% Generated - edit/create /tmp/.tmp_dir396237356/sys.config instead.
[{sasl,[{errlog_type,error}]},
 {distillery,
     [{config_providers,
          [{'Elixir.Mix.Releases.Config.Providers.Elixir',
               [<<"var/config.exs">>]}]}]},
 "extra.config"].
[{sasl,[{errlog_type,error}]},
 {distillery,
     [{config_providers,
          [{'Elixir.Mix.Releases.Config.Providers.Elixir',
               [<<"var/config.exs">>]}]}]},
 "extra.config"].
```

I believe I've seen the app not starting for some other reason, but I'm not able to reproduce it for now.

bitwalker:
I need to think about this one a bit more, I think there is a better way, but I need to finish some other changes I'm working on before I can circle back to this.

bitwalker:
Can you revisit master? Quite a bit has changed, and I suspect this is no longer a problem

studzien:
Yeah, looks good.
Perhaps the second commit still might makes sense, as I believe it might fix some of the travis failures?

bitwalker:
I've read through the second commit a few times, but it's not clear to me what it effectively changes - I'm sure I must be overlooking something. It looks like it ultimately swaps the `receive` timeout for an explicit `:timer.sleep`, but that doesn't actually address the race condition you found. It seems more likely that what we want to do is have an initial sleep of 1s before we start waiting, since that should allow the start command to create the `var/sys.config` file without the `ping` command interfering. Does that sound right to you?

bitwalker:
Yeah, definitely agreed on the non-determinism - I hate `sleep` anywhere in tests as it almost always will fail randomly at some point, better to find a way to wait and guarantee execution order somehow. I've been trying to fix up the tests in master, but I still get occasional failures - I can't reproduce locally, but since they fail in Travis regularly, I'm certain there is something I haven't pinned down yet.

PR #418, opened 2018-05-27
Fix generating a release upgrade from already existing appup

studzien:
### Summary of changes

It looks like generating a release from already provided appup files (including ones generated by `mix release.gen.appup`) is not possible because of usage of `and` instead of `&&`. 

Very small change, but it seemed untested, so added a test as well.
I tried to move some of the integration test logic, so there's less repetition, but I'm not sure if that's the way you'd like to go.

### Checklist


bitwalker:
Thanks! I think refactoring the tests is needed, so I've merged this as-is, but I think we probably need to go a step further to make individual steps more reusable - but this will work for now!

PR #420, opened 2018-06-04
Be more specific about where to include config for migration script

jsparling:
### Summary of changes

Although somewhat intuitive, it was unclear to me where the migration script configuration should go. This PR makes that location explicit.

FYI: the docs on the documentation site do not align with the repo, the repo ReleaseTasks example code works, but the sample code from the doc site does not. 

### Checklist


bitwalker:
Thanks!

PR #430, opened 2018-06-29
Added Remote Interactive to Logged output

smorin:
Added Remote Interactive to Logged output

### Summary of changes

Wanted to see the command for remote_console added to the list since the alternatives are listed.

### Checklist


bitwalker:
I'm not sure I'm +1 on this - if a user tries to run that command at the prompt right after the message, it will fail since the release is not running; the purpose of that message is to give new users a chance to see how to try out their newly created release. Thoughts?

smorin:
Think a comment can be added to the output.

# connect to a running release

bitwalker:
I've reworked the entirety of the help output for `mix release`, along with the success output printed when a build completes, it provides a more complete introduction. I'm closing this in favor of those changes, but thanks for the PR and the suggestion!

PR #434, opened 2018-07-27
print the custom command without filepath

tverlaan:
### Summary of changes

The current version prints the full path for custom commands. This strips the path and removes the `.sh` suffix if there is one. This should make it easier to directly see the commands.

### Checklist


tverlaan:
I'd also like to see this land on 1.5.x if possible.

PR #439, opened 2018-07-31
Encapsulate { into [{] so it is not interpreted as 'grouping'

4eek:
### Summary of changes

Extended Mode uses { and } as special characters. Making grep implementations other then GNU grep fail.

From http://www.gnu.org/software/grep/manual/grep.html#Basic-vs-Extended ...

```
3.6 Basic vs Extended Regular Expressions

In basic regular expressions the meta-characters ‘?’, ‘+’, ‘{’, ‘|’, ‘(’, and ‘)’ lose their special meaning; instead use the backslashed versions ‘?’, ‘+’, ‘{’, ‘|’, ‘(’, and ‘)’.

Traditional egrep did not support the ‘{’ meta-character, and some egrep implementations support ‘{’ instead, so portable scripts should avoid ‘{’ in ‘grep -E’ patterns and should use ‘[{]’ to match a literal ‘{’.

GNU grep -E attempts to support traditional usage by assuming that ‘{’ is not special if it would be the start of an invalid interval specification. For example, the command ‘grep -E '{1'’ searches for the two-character string ‘{1’ instead of reporting a syntax error in the regular expression. POSIX allows this behavior as an extension, but portable scripts should avoid it.
```

As per "**so portable scripts should avoid ‘{’ in ‘grep -E’ patterns and should use ‘[{]’ to match a literal ‘{’.**" I have updated the regex encapsulating { into [{] so it is not interpreted as 'grouping'.

### Checklist

### DIFF
--- a/priv/libexec/erts.sh
+++ b/priv/libexec/erts.sh
@@ -5,7 +5,7 @@ set -e
 
 code_paths() {
     __rel="$RELEASE_ROOT_DIR/releases/$REL_VSN/$REL_NAME.rel"
-    grep -E '{[A-Za-z_]*,\"[0-9.]*[A-Za-z0-9.\_\-\+]*\"(,[a-z]*)?}' "$__rel" \
+    grep -E '[{][A-Za-z_]*,\"[0-9.]*[A-Za-z0-9.\_\-\+]*\"(,[a-z]*)?[}]' "$__rel" \
         | grep -v "erts" \
         | sed -e's/"[^"]*$//' \
               -e's/^[^a-z]*//' \


4eek:
Tested on macOS...

```
$ grep -E '[{][A-Za-z_]*,\"[0-9.]*[A-Za-z0-9.\_\-\+]*\"(,[a-z]*)?[}]' "_build/dev/rel/acme_web/releases/0.0.1/acme_web.rel"

         {erts,"10.0.4"},
         [{kernel,"6.0"},
          {stdlib,"3.5.1"},
          {compiler,"7.2.2"},
          {elixir,"1.7.1"},
          {iex,"1.7.1"},
          {sasl,"3.2"},
          {runtime_tools,"1.13"},
          {mix,"1.7.1",load},
          {logger,"1.7.1"},
          {artificery,"0.2.5"},
          {distillery,"2.0.0"},
          {gettext,"0.15.0"},
          {crypto,"4.3"},
          {phoenix_pubsub,"1.0.2"},
          {public_key,"1.6.1"},
          {ssl,"9.0"},
          {ranch,"1.3.2"},
          {cowlib,"1.0.2"},
          {cowboy,"1.1.2"},
          {mime,"1.3.0"},
          {plug,"1.6.0"},
          {phoenix_html,"2.11.2"},
          {poison,"3.1.0"},
          {eex,"1.7.1"},
          {phoenix,"1.3.3"},
          {file_system,"0.2.5"},
          {phoenix_live_reload,"1.1.5"},
          {elixir_uuid,"1.2.0"},
          {elixir_make,"0.4.2"},
          {comeonin,"4.1.1"},
          {jose,"1.8.4"},
          {joken,"1.5.0"},
          {acme_web,"0.0.1"}]}.
```

4eek:
Tested on Alpine:latest...

```
bash-4.4# busybox grep -E '[{][A-Za-z_]*,\"[0-9.]*[A-Za-z0-9.\_\-\+]*\"(,[a-z]*)?[}]' "/acme_web/releases/0.0.1/acme_web.rel"

         {erts,"10.0.3"},
         [{kernel,"6.0"},
          {stdlib,"3.5.1"},
          {compiler,"7.2.2"},
          {elixir,"1.7.0"},
          {iex,"1.7.0"},
          {sasl,"3.2"},
          {runtime_tools,"1.13"},
          {mix,"1.7.0",load},
          {logger,"1.7.0"},
          {artificery,"0.2.5"},
          {distillery,"2.0.0"},
          {gettext,"0.15.0"},
          {crypto,"4.3"},
          {phoenix_pubsub,"1.0.2"},
          {public_key,"1.6.1"},
          {ssl,"9.0"},
          {ranch,"1.3.2"},
          {cowlib,"1.0.2"},
          {cowboy,"1.1.2"},
          {mime,"1.3.0"},
          {plug,"1.6.0"},
          {phoenix_html,"2.11.2"},
          {poison,"3.1.0"},
          {eex,"1.7.0"},
          {phoenix,"1.3.3"},
          {elixir_uuid,"1.2.0"},
          {elixir_make,"0.4.2"},
          {comeonin,"4.1.1"},
          {jose,"1.8.4"},
          {joken,"1.5.0"},
          {acme_web,"0.0.1"}]}.
```

4eek:
Created issue #440 for this PR.

4eek:
It's been suggested that installing _proper_ grep on the Alpine image with `RUN apk --update add grep` is a solution. This does work but I feel the cleaner solution is a portable grep command rather than adding to the Alpine base image.

bitwalker:
Thanks!

PR #448, opened 2018-08-02
Fix equality check of MapSet against List

imranismail:
### Summary of changes

Fix unnecessary warning for loaded apps

### Checklist


bitwalker:
I ended up applying a fix by hand with a slight adjustment, thanks for the PR though!

PR #450, opened 2018-08-02
Fix pattern matching for ERTS version

JustMikey:
In case of success, `detect_erts_version/1` returns a tuple `{:ok, vsn}` instead of just the version value. This PR fixes this small issue.


bitwalker:
I had to apply this change manually since I just pushed a bunch of changes.

PR #451, opened 2018-08-02
Updates to allow things to be silenced

mobileoverlord:
Why?
`mix release --silent` should silence all output. This PR updates the final message to go through `Logger.info` instead of `IO.puts`

`mix release.clean` Adds the switch `--silent` to set the verbosity level.


bitwalker:
I've had to manually apply these changes, since I literally _just_ rewrote a bunch of that stuff haha.

PR #452, opened 2018-08-02
Fix issues with plugins

mobileoverlord:
This PR does the following:
* Fixes issue with using `not` to check `include_erts` when it is a string path.
* Fixes issue were plugin modules are not called because the modules code is not loaded.

### DIFF
--- a/priv/templates/release_rc_exec.eex
+++ b/priv/templates/release_rc_exec.eex
@@ -11,7 +11,7 @@ RELEASES_DIR="$RELEASE_ROOT_DIR/releases"
 REL_NAME="${REL_NAME:-<%= release_name %>}"
 REL_VSN="${REL_VSN:-$(cut -d' ' -f2 "$RELEASES_DIR"/start_erl.data)}"
 ERTS_VSN="${ERTS_VSN:-$(cut -d' ' -f1 "$RELEASES_DIR"/start_erl.data)}"
-<%= if (not include_erts or is_nil(erts_vsn)) do %>USE_HOST_ERTS=true<% end %>
+<%= if (include_erts == false or is_nil(erts_vsn)) do %>USE_HOST_ERTS=true<% end %>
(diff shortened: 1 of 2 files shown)


lib/mix/lib/releases/plugins/plugin.ex
```diff
@@ -169,10 +169,12 @@ defmodule Mix.Releases.Plugin do
     plugins =
       for p <- release.profile.plugins do
         case p do
-          {_p, _opts} ->
+          {mod, _opts} ->
+            load(mod)
```
bitwalker: I think a better place for this is in the config file - so that we load all plugin modules in one go, rather than hitting the code server for every event which hits all of the plugins. Not super important in Distillery, but it also guarantees that we can count on plugins being available everywhere after that.

priv/templates/release_rc_exec.eex
```diff
@@ -11,7 +11,7 @@ RELEASES_DIR="$RELEASE_ROOT_DIR/releases"
 REL_NAME="${REL_NAME:-<%= release_name %>}"
 REL_VSN="${REL_VSN:-$(cut -d' ' -f2 "$RELEASES_DIR"/start_erl.data)}"
 ERTS_VSN="${ERTS_VSN:-$(cut -d' ' -f1 "$RELEASES_DIR"/start_erl.data)}"
-<%= if (not include_erts or is_nil(erts_vsn)) do %>USE_HOST_ERTS=true<% end %>
+<%= if (include_erts == "" or include_erts == false or is_nil(erts_vsn)) do %>USE_HOST_ERTS=true<% end %>
```
bitwalker: Were you seeing `include_erts` come up as an empty string? That should have triggered an error when ERTS is validated during `mix release`

mobileoverlord:
Simplified it with your comments. Rebased

PR #454, opened 2018-08-02
Only copy apps from erts path if the app is part of erts

mobileoverlord:
I am open to structuring this differently :)

### DIFF
--- a/lib/mix/lib/releases/assembler.ex
+++ b/lib/mix/lib/releases/assembler.ex
@@ -135,7 +135,13 @@ defmodule Mix.Releases.Assembler do
         copy_app(app_dir, target_dir, dev_mode?, include_src?)
 
       p when is_binary(p) ->
-        app_dir = Path.join([p, "lib", "#{app_name}-#{app_version}"])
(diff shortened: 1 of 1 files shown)


lib/mix/lib/releases/assembler.ex
```diff
@@ -160,6 +166,7 @@ defmodule Mix.Releases.Assembler do
   end
 
   defp copy_app(app_dir, target_dir, false, include_src?) do
+    IO.inspect app_dir
```
bitwalker: You forgot something :P

bitwalker:
I'm +1, just need to remove that inspect :)

lib/mix/lib/releases/assembler.ex
mobileoverlord: 😇

mobileoverlord:
Rebased, Whoops :)

mobileoverlord:
blaergh, stupid space

mobileoverlord:
Okay, _this_ time!

PR #455, opened 2018-08-02
Expand the path to vm_args

mobileoverlord:


lib/mix/lib/releases/assembler.ex
```diff
@@ -603,6 +603,7 @@ defmodule Mix.Releases.Assembler do
 
   defp generate_vm_args(%Release{profile: %Profile{vm_args: path}} = rel) do
     Shell.debug("Generating vm.args from #{Path.relative_to_cwd(path)}")
+    path = Path.expand(path)
```
bitwalker: We should probably move this to the config handler, since we'll apply the same expansion to all paths where we'll be using as a source for a copy or link.

bitwalker: I think actually I'll do that in a different commit, this is good for now

PR #463, opened 2018-08-06
Merge runtime config with compiled config

imranismail:
### Summary of changes

I just tried the latest change and right now the conflict resolution is not recursive like how mix is by default and users might not expect this to be the case

This PR fixes that

```elixir
iex(1)> Mix.Config.merge([app: [url: "https://app.com"]], [app: [port: 443]])
[app: [url: "https://app.com", port: 443]]
iex(2)> Keyword.merge([app: [url: "https://app.com"]], [app: [port: 443]])
[app: [port: 443]]
```

### Checklist


bitwalker:
Thanks!

PR #472, opened 2018-08-10
add guide to build a docker image with your release

happysalada:
### Summary of changes

added a guide to build your release in a docker image.
Ideally, this is the first step when you want to deploy with Docker or Kubernetes.

### Comments

- I don't really like the name of the doc file, but I couldn't find a better one
- let me know if you think this is a good first step and/or a good way of breaking things down
- once we agree that the content is good, I'll squash my commits

### DIFF
--- a/docs/guides/docker_image_with_your_release.md
+++ b/docs/guides/docker_image_with_your_release.md
@@ -49,9 +49,23 @@ COPY --from=0 /opt/release .
 CMD ["/opt/app/bin/start_app", "foreground"]
 ```
 
-verify that you can build your image with the following command
-`docker build -t my_tag --build-arg APP_VERSION="0.0.1" .`
-(it's common to tag images with your docker-hub username and repo name like `-t my_username/my_repo_name:version_number`)
+Verify that you can build your image
+
+Add the following `makefile` to your project
+
+```makefile
+BUILD ?= `git describe --always`
+IMAGE ?= `grep 'app:' mix.exs | sed -e 's/ //g' -e 's/app://' -e 's/[:,]//g'`
+VERSION ?= `grep 'version' mix.exs | sed -e 's/ //g' -e 's/version://' -e 's/[",]//g'`
+
+image:
+  docker build -t $(IMAGE):$(VERSION)-$(BUILD) .
+```
+
+Run `make` to build your image
+
+!!!warning
+    If you get the following error when running the makefile `makefile:6: *** multiple target patterns.  Stop.`, verify that the `docker build` has a tab and not spaces in front of it.
 
 !!!info
     Yarn is optional, it is used instead of npm to reduce dependency fetching by a consequent amount of time.
--- a/docs/guides/docker_image_with_your_release.md
+++ b/docs/guides/docker_image_with_your_release.md
@@ -1,23 +1,12 @@
 # Build a docker image with your release
 
-Now that we have seen how to build a release, let's build one in a docker image. This is necessary for a deployment with docker swarm or kubernetes for example.
+## Build the image
 
-A prerequisite step to this, is to initialize your release (`mix release.init`) and configure your [vm.args](/files/vm.args).
+Now that we have seen how to build a release, let's build one in a docker image. This is necessary for a deployment with Docker Swarm or Kubernetes for example.
 
-in order to customize the application release version when building the docker image, change the following in your `mix.exs` file
+Add a `dockerfile` to your project with the following content
 
-```
-# add this
-@version System.get_env("APP_VERSION") || "0.0.0"
-
-def project do
-  app: :my_app,
-  version: @version, # change this line
-```
-
-once that is done add a `dockerfile` to your project with the following content
-
-```
+```docker
 # check https://hub.docker.com/_/elixir/ for the appropriate version of elixir you want to run
 # alpine is used here to make dependency fetching faster
 FROM elixir:1.7.2-alpine
@@ -25,18 +14,13 @@ FROM elixir:1.7.2-alpine
 # these arg command, set a default for the env variables
 ARG APP_NAME=your_app_name
 ARG MIX_ENV=prod
-ARG APP_VERSION=0.0.0
 # use this variable in case you use an umbrella project
 ARG PHOENIX_SUBDIR=.
 ENV MIX_ENV ${MIX_ENV}
-ENV APP_VERSION ${APP_VERSION}
-# We choose to handle environment variables in our application by replacing System.get_env("MY_VAR")
-# with "${MY_VAR}". Setting up REPLACE_OS_VARS to true, will replace the values for us.
 ENV REPLACE_OS_VARS true
 
 # in linux the opt dir is used to put things not directly required by the os
 WORKDIR /opt/app
-# use yarn instead of npm to reduce dependency fetching by a lot (from 180s to 60s on my machine on my project)
 # add git if one of your dependency pulls from git
 RUN apk update \
   && apk --no-cache --update add nodejs yarn git build-base \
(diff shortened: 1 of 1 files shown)


docs/guides/docker_image_with_your_release.md
```diff
@@ -0,0 +1,58 @@
...
+
+RUN cd ${PHOENIX_SUBDIR}/assets \
+  && yarn install \
+  && yarn deploy \
+  && cd .. \
+  && mix phx.digest
+# we change the name of the start script here, just so this file can be copied and pasted
+RUN mix release --env=${MIX_ENV} --verbose \
+  && mv _build/prod/rel/${APP_NAME} /opt/release \
+  && mv /opt/release/bin/${APP_NAME} /opt/release/bin/start_app
```
happysalada: Here the start script name is changed, because the CMD docker part doesn't take environment variables.
Given that your app name won't likely change, it is just a matter of renaming `start_app` everywhere with `your_app_name`.  I know this is debatable, let me know if you feel strongly either way.

docs/guides/docker_image_with_your_release.md
```diff
@@ -0,0 +1,58 @@
...
+ARG APP_NAME=your_app_name
+ARG MIX_ENV=prod
+ARG APP_VERSION=0.0.0
+# use this variable in case you use an umbrella project
+ARG PHOENIX_SUBDIR=.
+ENV MIX_ENV ${MIX_ENV}
+ENV APP_VERSION ${APP_VERSION}
+# We choose to handle environment variables in our application by replacing System.get_env("MY_VAR")
+# with "${MY_VAR}". Setting up REPLACE_OS_VARS to true, will replace the values for us.
+ENV REPLACE_OS_VARS true
```
happysalada: I use this "simple" way for dealing with env vars, this might potentially change for 2.0. Let me know if you want to change it

docs/guides/docker_image_with_your_release.md
```diff
@@ -0,0 +1,58 @@
...
+
+```
+# check https://hub.docker.com/_/elixir/ for the appropriate version of elixir you want to run
+# alpine is used here to make dependency fetching faster
+FROM elixir:1.7.2-alpine
+
+# these arg command, set a default for the env variables
+ARG APP_NAME=your_app_name
+ARG MIX_ENV=prod
+ARG APP_VERSION=0.0.0
```
happysalada: I found this in some repo from the financial times, I thought that customizing version from when building the image was convenient. Let me know if you don't like it.

docs/guides/docker_image_with_your_release.md
```diff
@@ -0,0 +1,74 @@
...
+  && mv /opt/release/bin/${APP_NAME} /opt/release/bin/start_app
+
+# minimal runtime image
+FROM alpine:3.8
+# bash is required by distillery
+RUN apk update && apk --no-cache --update add bash openssl-dev
+ENV REPLACE_OS_VARS true
+WORKDIR /opt/app
+COPY --from=0 /opt/release .
+CMD ["/opt/app/bin/start_app", "foreground"]
```
happysalada: I debatted (with myself) splitting this into an ENTRYPOINT and a CMD like
```
ENTRYPOINT ["/opt/app/bin/start_app"]
CMD ["foreground"]
```

the disadvantage of that is that you wouldn't be able to override completely the start_app script to run something completely different on this image. However would you ever want to do this ever?
the advantage is that when you need to run `console_clean` you just need to add that to your docker command and not the whole path of where the start_app script is located (which if you don't remember can be a pain)

archSeer:
I've found bitwalker's Dockerfiles a lot more useful:

- https://github.com/bitwalker/alpine-elixir
- https://github.com/bitwalker/alpine-erlang

Use alpine-elixir to build a distillery release tarball, then take that tarball and put it in alpine-erlang (or even better, we include erts so we build a custom alpine image with just the required erts dependencies for a smaller image size)

You could do both steps in one file with multistage builds:

https://docs.docker.com/develop/develop-images/multistage-build/

docs/guides/docker_image_with_your_release.md
```diff
@@ -0,0 +1,74 @@
...
+
+Now that we have seen how to build a release, let's build one in a docker image. This is necessary for a deployment with docker swarm or kubernetes for example.
+
+A prerequisite step to this, is to initialize your release (`mix release.init`) and configure your [vm.args](/files/vm.args).
+
+in order to customize the application release version when building the docker image, change the following in your `mix.exs` file
+
+```
+# add this
+@version System.get_env("APP_VERSION") || "0.0.0"
```
bitwalker: In general I find it easier to use the `:version` field in `mix.exs` normally, and grep it out of `mix.exs` in a Makefile which is used to build images, i.e.:

```mix.exs
def project do
  # ...
  version: "0.1.0"
```

```make
IMAGE ?= `grep 'app:' mix.exs | sed -e 's/ //g' -e 's/app://' -e 's/[:,]//g'`
VERSION ?= `grep 'version' mix.exs | sed -e 's/ //g' -e 's/version://' -e 's/[",]//g'`

#...
image: ## Build Docker image
    docker build -t $(IMAGE):$(VERSION) .
```

This is portable to any project, and will build a tagged, versioned image of the form `myapp:0.1.0` by running `make image` in the project root.

happysalada: thanks for the tip with grep and sed.
What I liked about this, is that it doesn't force you to commit a new version number everytime you make changes to the code. I feel almost that the version number is a config and shouldn't appear in the code. I understand it's a matter of taste.

I'll replace this with this new way then.

docs/guides/docker_image_with_your_release.md
```diff
@@ -0,0 +1,74 @@
...
+  && yarn deploy \
+  && cd .. \
+  && mix phx.digest
+# we change the name of the start script here, just so this file can be copied and pasted
+RUN mix release --env=${MIX_ENV} --verbose \
+  && mv _build/prod/rel/${APP_NAME} /opt/release \
+  && mv /opt/release/bin/${APP_NAME} /opt/release/bin/start_app
+
+# minimal runtime image
+FROM alpine:3.8
```
happysalada: @archSeer I'm open to feedback, however I don't understand your point.
This docker image is using multi-stage build (from this line). Maybe it's not clear enough?

I like what bitwalker does in general, but I thought, that for people who don't know him, it would be better to use the official image. I'm not sure what using a different image would change here. If you can maybe clarify, that would be helpful.

bitwalker: I'm fine with using a different image than the ones I build, as long as it is an "official" image, otherwise it's easier for me to maintain the guides with images I'm using. So you can leave it as-is :)

docs/guides/docker_image_with_your_release.md
bitwalker: My take on that is that you should append a git shorthash to the version number if you are building new images while not changing the version - this keeps the application version the same, but the release artifact gets a distinct "release" version:

```make
BUILD ?= `git describe --always`
IMAGE ?= `grep 'app:' mix.exs | sed -e 's/ //g' -e 's/app://' -e 's/[:,]//g'`
VERSION ?= `grep 'version' mix.exs | sed -e 's/ //g' -e 's/version://' -e 's/[",]//g'`

image: ## Build Docker image
    docker build -t $(IMAGE):$(VERSION)-$(BUILD)
```

docs/guides/docker_image_with_your_release.md
bitwalker: I think using just `CMD` is preferable - I will often pop into an image to run other tasks if doing some troubleshooting, and so the entrypoint is not always going to be the release script.

docs/guides/docker_image_with_your_release.md
```diff
@@ -0,0 +1,74 @@
+# Build a docker image with your release
+
+Now that we have seen how to build a release, let's build one in a docker image. This is necessary for a deployment with docker swarm or kubernetes for example.
```
bitwalker: Make sure to capitalize all the pronouns

docs/guides/docker_image_with_your_release.md
```diff
@@ -0,0 +1,74 @@
...
+
+A prerequisite step to this, is to initialize your release (`mix release.init`) and configure your [vm.args](/files/vm.args).
```
bitwalker: I would make this a callout, e.g.

```
!!! info
    Make sure you have initialized Distillery in your project with `mix release.init` and configured your `vm.args` file as mentioned [here](../config/vm.args.md)
```

Do they need to configure `vm.args` though? If it's not actually a prerequisite, I would leave that out.

docs/guides/docker_image_with_your_release.md
```diff
@@ -0,0 +1,74 @@
...
+```
+# add this
+@version System.get_env("APP_VERSION") || "0.0.0"
+
+def project do
+  app: :my_app,
+  version: @version, # change this line
+```
+
+once that is done add a `dockerfile` to your project with the following content
```
bitwalker: Make sure to capitalize sentences

docs/guides/docker_image_with_your_release.md
```diff
@@ -0,0 +1,74 @@
...
+WORKDIR /opt/app
+COPY --from=0 /opt/release .
+CMD ["/opt/app/bin/start_app", "foreground"]
+```
+
+verify that you can build your image with the following command
+`docker build -t my_tag --build-arg APP_VERSION="0.0.1" .`
+(it's common to tag images with your docker-hub username and repo name like `-t my_username/my_repo_name:version_number`)
+
+a couple of things to note:
```
bitwalker: Use callouts like I described above, either `!!! info` or `!!! tip`

happysalada:
@bitwalker ready for another round of review

docs/guides/docker_image_with_your_release.md
happysalada: One question regarding the BUILD variable, shouldn't it be `git rev-parse --short HEAD` ?

`git describe --always` gives me a full version number like `v0.0.1-355-g97f75e69`
(I have to admit, I'm not super familiar with makefiles, so maybe there is something I am missing)

my second question relates to the build stage of a continuous integration agent. It would look something like this then ?

```
BUILD= `git rev-parse --short HEAD`
            IMAGE= `grep 'app:' mix.exs | sed -e 's/ //g' -e 's/app://' -e 's/[:,]//g'`
            VERSION= `grep 'version' mix.exs | sed -e 's/ //g' -e 's/version://' -e 's/[",]//g'`
            make
            docker login -u $DOCKER_LOGIN -p $DOCKER_PASSWORD
            docker push dockerhub_username/$IMAGE:$VERSION-$BUILD
            docker push dockerhub_username/$IMAGE:latest
```

happysalada: for anybody else interested, make wasn't directly available on circle_ci machine so I ended going with the following
```
echo "export BUILD=$(git rev-parse --short HEAD)" >> $BASH_ENV
echo "export IMAGE=$(grep 'app:' mix.exs | sed -e 's/ //g' -e 's/app://' -e 's/[:,]//g')" >> $BASH_ENV
echo "export VERSION=$(grep 'version' mix.exs | sed -e 's/ //g' -e 's/version://' -e 's/[",]//g')" >> $BASH_ENV
source $BASH_ENV
docker build -t happysada/$IMAGE:$VERSION-$BUILD -t happysalada/$IMAGE:latest .
docker login -u $DOCKER_LOGIN -p $DOCKER_PASSWORD
docker push happysalada/$IMAGE:$VERSION-$BUILD
docker push happysalada/$IMAGE:latest
```

happysalada:
@beardedeagle thank you for your comment.
A couple of problems I see with your link are:
- phoenix is about to switch to webpack, that's why I don't explicitely use brunch, but go with yarn instead.
- The second problem is about the app name. I added it as an environment variable, so that anybody can just copy and paste that file. With your file, they would have to edit the app name everytime.

However, it's a very good file!
One of the good things is that you name your stages, perhaps I should do that as well.

beardedeagle:
Yes, until the new version of Phoenix releases, I will not be switching my docs to use webpack. Once it does though, I will add webpack in.

bitwalker:
Merged manually in 3f3acf6

I've made significant changes to the document though - can everyone review the final version and give me your feedback? If there are problems, please open a new issue and we can discuss them there. I think I'd like to take it a step further and build an example application like I do in the AWS guide, that way there is less hedging around the things that can change. People are smart enough to figure out what things apply to them or don't - but they usually want guidance on how to achieve specific goals. I'm less concerned with making a guide that is "copy-pasteable", and more concerned with making one that can be followed from beginning to finish successfully, with enough detail shown that a new user can remove the parts they don't need, or modify them for their use case.

Thanks again for all the help @happysalada!

happysalada:
The new guide looks good!
(didn't see any problem with it)

Issue #441, opened 2018-08-01
Improve Documentation


bitwalker:
This issue is to track the work for improving the documentation:

- [x] Split up guides/pages on topics to provide softer introductions with specific pages for more detailed content
- [ ] Provide intro page with basic rationale of releases, terminology, and starting points for different languages (i.e. I'm coming from Ruby or Python, here is how OTP releases compare).
- [ ] Provide doc to provide guidance on operations, troubleshooting, third-party tools
- [x] Provide "deploying with Docker" guide
- [x] Provide "deploying to Digital Ocean" guide (or anything which is just deploying to a VM)
- [x] Provide "deploying to AWS" guide
- [ ] Need documentation for Windows (HELP WANTED)

I'm open to suggestions for other guides/topics we should cover or address. I would prefer if you check with me first before opening PRs for documentation, as I'm going to be doing a lot of moving around and rewriting up front, and I would hate to conflict with anything you may want to help contribute. If we talk it out first, I can make sure we don't step on each others toes :)

happysalada:
I have written [this guide on deploying with docker](https://github.com/happysalada/phoenix-the-unoffcial-guide/blob/master/deployment/docker.md) before. I'm thinking a lot can be re-used, since that's what I use to deploy to Digital Ocean.
The only thing I couldn't solve was how to get the -heart option working to restart automatically in case of crash.
Are you looking for something different in terms of docker/digital ocean documentation? Or is the direction generally good? Let me know.

bitwalker:
@happysalada There is some information there that will need to be updated, but it's definitely a solid guide for Docker. If you are interested in making that part of the official documentation, a PR would be welcome! We can work through any modifications necessary to update it for 2.x, but they should be mostly minimal.

If you want to maintain that guide separately from Distillery's docs, I can write one up based on my own experience as well when I get some time.

> The only thing I couldn't solve was how to get the -heart option working to restart automatically in case of crash.

In Docker you mean? That would always be doomed to fail, because if the VM crashes and `heart` restarts it, it will have a new pid - but regardless, the original command to start the VM would have exited on the original crash, so the container would die anyway. In a Docker environment you are better off letting the container die and have your orchestration layer deal with restarting it, I don't think `heart` is a good fit there - it is better suited for running a release as a daemon on a system with no supervision (i.e. without something like systemd or supervisord)

happysalada:
[...]

PR #473, opened 2018-08-10
add docker_deployment guide

happysalada:
### Summary of changes

add a deploy with docker guide related to #441 

### Comments
- should we add a part on how to do migrations here?
- I thought about documenting how you do upgrade in the digitial ocean guide since there are a couple of unintuitive things there.  Should I still talk about it briefly here?
- are you ok with the split that docker deployment uses docker swarm but deploys on your machine. Then in another guide we treat how to deploy on other machines.

### DIFF
--- a/docs/guides/docker_deployment.md
+++ b/docs/guides/docker_deployment.md
@@ -51,10 +51,6 @@ if you have external dependencies this is the place to add them. For example if 
 ```
 version: '3.5'
 
-volumes:
-    prometheus_data: {}
-    grafana_data: {}
-
 networks:
   webnet:
     driver: overlay
--- a/docs/guides/docker_deployment.md
+++ b/docs/guides/docker_deployment.md
@@ -1,0 +1,109 @@
+# Docker deployment
+
+In this guide we will cover how to deploy with docker. We will deploy on our local machine, just in the same fashion that you should deploy in production. However to deploy on a particular machine, commands will change slightly. (for these commands for a specific provider, check the specific provider section, for example digital ocean)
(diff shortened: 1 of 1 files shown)


docs/guides/docker_deployment.md
```diff
@@ -0,0 +1,109 @@
...
+    networks:
+      - webnet
+
+# careful what you name your service here, these will be your database hostname in your env vars
+  db:
+    image: postgres:10-alpine
+    deploy:
+      replicas: 1
+      placement:
+        constraints: [node.role == manager]
```
happysalada: I am not 100% sure about this `placement: constraints` thing. From what I understand you need to put this if you use the file system. Since you want to persist your postgres data, I thought it was necessary.

docs/guides/docker_deployment.md
```diff
@@ -0,0 +1,105 @@
...
+if you have external dependencies this is the place to add them. For example if you have a database, change your file to the following
+
+```
+version: '3.5'
+
+networks:
+  webnet:
+    driver: overlay
+    attachable: true
+
```
happysalada: some people declare their volumes before the services in their docker-compose app. I'm not really sure why/why that is needed.

bitwalker:
I think I would prefer to make this part of the other Docker document; sort of a "you've built an image, let's run it locally with Docker Compose to show how you can run multiple images together". I would prefer "deploying with Docker" guides be tailored towards specific environments (ECS, Kubernetes in it's various forms, etc.)

PR #481, opened 2018-08-13
Problem: upfrom option is not recognized

lessless:
In 1.5.x `upfrom` option was merged with defaults https://github.com/bitwalker/distillery/blob/1.5.x/lib/distillery/tasks/release.ex#L217
and override `:latest` default. Probably its parsing was lost during 2.x upgrade.

### Summary of changes

### Checklist


bitwalker:
Thanks!

PR #493, opened 2018-08-18
fix working_with_docker doc

happysalada:
### Summary of changes

Three things
- There is a typo where REL_VSN is used, but it's never defined
- I found that defining ARG before build doesn't work. They need to be defined in each stage they are used. (I didn't know that, [the official doc is a little ambiguous though](https://docs.docker.com/engine/reference/builder/#scope))
- pushd and popd are not available. Probably because it's an alpine image.

### DIFF
--- a/docs/guides/working_with_docker.md
+++ b/docs/guides/working_with_docker.md
@@ -20,6 +20,13 @@ In the root of your project, create a new file named `Dockerfile` with the
 following content:
 
 ```docker
+# The version of Alpine to use for the final image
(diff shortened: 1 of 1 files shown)


bitwalker:
I ended up reworking the document by hand by just walking through the guide using my example application `distillery-test`. I haven't done the docker-compose parts yet, but the image build/run works as expected. I'm going to create an example app for this guide, I just haven't gotten around to it yet.

Thanks for reporting the issues!

PR #494, opened 2018-08-18
Update link to deployment docs

zacck:
I don't think the Advanced deployment link still exists changing to a link that describes phoenix release setups

### Summary of changes

### Checklist

N/A


bitwalker:
Good call, thanks!

PR #499, opened 2018-08-22
Ensure docker.env file matches config.exs

rob0t7:
The current working_with_docker.md file contains a mistake when setting production configuration variables. 

There seems to exist a different name convention between **config/docker.env** and **rel/config/config.exs**. This fixes the documentation to ensure that the tutorial works if followed properly.

### Summary of changes

### Checklist


bitwalker:
Thanks :) - I added my test project as an example after the fact, and never got around to updating that file in the guide - much appreciated!

PR #501, opened 2018-08-23
About stopping "simple" systemd service

kluufger:
👋

### Summary of changes

Is `ExecStop=/home/appuser/myapp/bin/myapp stop` really required for `Type=simple` service?

### Checklist

### Licensing/Copyright


kluufger:
At least the AWS [demo](https://github.com/bitwalker/distillery-aws-example/blob/master/rel/etc/distillery-example.service) doesn't seem to have it.

bitwalker:
You're right, it isn't needed, by default it will send SIGTERM when the service needs to stop, which is handled by OTP's signal handler properly since OTP 20.

PR #506, opened 2018-08-25
Fix empty tip box in 'deploying to aws' guide

sabondano:
### Summary of changes

I noticed an empty tip box. I think it's because of how it's indented, so here's the edit accordingly. Just a guess though, never used mkdocs before 😊
screen shot 2018-08-25 at 9 59 07 am

### Licensing/Copyright


bitwalker:
Good catch!

PR #511, opened 2018-08-27
Convert --profile args to atoms

aquarhead:
### Summary of changes

Currently --profile switch doesn't work, after adding some debug log and inspecting, I figured out it's because the --profile args aren't converted to atoms so `Map.get` can't find the specified environment (and release name).

with `--profile`:

```
  selected_environment: "prod",
  selected_release: "xyz",
```

with `--env`:

```
  selected_environment: :prod,
  selected_release: :default,
```

### Licensing/Copyright


bitwalker:
Thanks, the arg handling changed in 2.x, so this one got missed. Thanks for the PR!

PR #518, opened 2018-08-29
Spurious code in migration; path quoting pidfile

DianaOlympos:
### Summary of changes

- Quoting of pidfile path need to use both quotes.
- Deleting the code to check the application in the Migration example, as they are loaded already

### Licensing/Copyright


bitwalker:
Thanks!

PR #522, opened 2018-08-31
Fix 519

### LINKED ISSUE #519: Release command takes forever
### Steps to reproduce

`mix release` under dev or prod

I was at 2.0.5 and after updating to 2.0.9 this happens.

### Description of issue

- What are the expected results?
Shouldn't take too long
- What version of Distillery?
2.0.9
- What OS, Erlang/Elixir versions are you seeing this issue on?
Debian and MacOS

Schultzer:
### Summary of changes

Prevent unnecessary recursive loop.

When moving the `:digraph.vertex/2` Check to `add_app/4` we can prevent an unnecessary recursive loop there happens in umbrella projects where we go through the same dependencies even when we already have seen them.

this fixes #519.

### Checklist

### Licensing/Copyright

### DIFF
--- a/lib/mix/lib/releases/models/release.ex
+++ b/lib/mix/lib/releases/models/release.ex
@@ -508,39 +508,36 @@ defmodule Mix.Releases.Release do
     add_apps(dg, as, apps)
   end
 
-  defp add_app(dg, as, parent, app)
-
   defp add_app(dg, as, parent, {name, start_type}) do
-    do_add_app(dg, as, parent, App.new(name, start_type))
-  end
-
-  defp add_app(dg, as, parent, name) do
-    do_add_app(dg, as, parent, App.new(name))
-  end
-
-  defp do_add_app(dg, as, parent, app) do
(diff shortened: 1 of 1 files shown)


lib/mix/lib/releases/models/app.ex
```diff
@@ -22,14 +22,7 @@ defmodule Mix.Releases.App do
   @valid_start_types [:permanent, :temporary, :transient, :load, :none]
 
   @doc """
-  Create a new Application struct from an application name
-  """
-  @spec new(atom) :: nil | __MODULE__.t() | {:error, String.t()}
```
bitwalker: Can you add this function back? It's part of the public API and isn't relevant to the fix for the issue this PR targets

bitwalker:
Overall this looks like it should be the right fix, other than the one review comment I've left. I'll merge this once you've updated the PR. Thanks for looking into the issue, and for the contribution!

Schultzer:
Done

bitwalker:
Thanks!

PR #527, opened 2018-09-05
FIX Dockerfile.alpine for docker build

peccu:
### Summary of changes

Alpine container does not have pushd, popd and bash.
I removed them.
WORKDIR is set to `/opt` so I think we don't need pushd/popd.

### Licensing/Copyright

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -9,8 +9,6 @@ RUN \
   mkdir -p /opt/distillery-test && \
   mix local.rebar --force && \
   mix local.hex --force && \
-  pushd /opt && \
-  git clone https://github.com/bitwalker/distillery-test && \
-  popd
+  git clone https://github.com/bitwalker/distillery-test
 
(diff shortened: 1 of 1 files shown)


Dockerfile.alpine
```diff
@@ -9,8 +9,6 @@ RUN \
   mkdir -p /opt/distillery-test && \
   mix local.rebar --force && \
   mix local.hex --force && \
-  pushd /opt && \
-  git clone https://github.com/bitwalker/distillery-test && \
-  popd
+  git clone https://github.com/bitwalker/distillery-test
 
-CMD ["bin/bash"]
```
bitwalker: Can you change this back, `bash` is required for Distillery, and I prefer using it by default. This image is for my own local testing anyway. Other than that 👍

peccu: Thank you for your review.
I fixed it to full path to `bash`.

Issue #540, opened 2018-09-19
Application not starting on Windows


fmooreRC:
### Steps to reproduce

The build completes successfully, but when I try to run the application on the Windows server, it results in an error.  I followed the walkthrough documentation [here](https://hexdocs.pm/distillery/introduction/walkthrough.html).  I've tried using different versions of Distillery and ended up with different results.

### Verbose Logs

Distillery 2.0.8
```
C:\opt\app> bin/app start
{“init terminating in do_boot”,{‘cannot get bootfile’,‘.boot’}}
init terminating in do_boot ({cannot get bootfile,.boot})

Crash dump is being written to: erl_crash.dump...done
Files\erl10.0.1\bin\erl.exe”“==“” was unexpected at this time.
```

Distillery 2.0.10
```
C:\opt\app> bin/app start
eq was unexpected at this time.
```

### Description of issue

- What are the expected results?
The app should start and run on the Windows server.

- What version of Distillery?
2.0.8 and 2.0.10

- What OS, Erlang/Elixir versions are you seeing this issue on?

Build is occurring on Mac OS X
Erlang/OTP 21 [erts-10.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]
Elixir 1.6.5 (compiled with OTP 20)

Release is being deployed to Windows Server 2014:
Erlang/OTP 21 [erts-10.0.1] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1]
Elixir 1.7.3 (compiled with Erlang/OTP 19)

- If possible, also provide your `rel/config.exs`, as it is often
  my first troubleshooting question, and you'll save us both time :)
I've copied the compiled windows erts files to my Mac.  Here's the relevant part of my rel/config.exs file:
```
~w(rel plugins *.exs)
|> Path.join()
|> Path.wildcard()
|> Enum.map(&Code.eval_file(&1))

use Mix.Releases.Config,

environment :dev_server do
  set include_erts: "../erl10.0.1"
  set include_src: false
  set cookie: :redacted
end

release :fergus do
  set version: current_version(:fergus)
  set applications: [
    :runtime_tools
  ]
end

```
- Is there documentation that says one thing, but Distillery does
  another? If so, please link the doc here so it can be updated if
  it's a documentation issue, or so that the fix can be based around
  what's documented.
- If this is a runtime configuration issue, please also provide your config file 
  (with any sensitive information stripped of course). This is almost
[...]

PR #541, opened 2018-09-19
Fix the equality comparison in windows batch files

fdbeirao:
According to [documentation](https://ss64.com/nt/if.html), equality comparison in Windows batch files should use the [`equ` operator](https://ss64.com/nt/equ.html).

The previous version would output `eq was unexpected at this time.`.

This PR addresses issue #540 .

### Licensing/Copyright


bitwalker:
Thanks!

PR #555, opened 2018-10-01
clean spurious whitespace in mix.exs

DianaOlympos:
### Summary of changes
To make the `vm.args` file more explicit to the user, we create a default `vm.args` that we put at the root of the `rel` folder through `release.init`. Additionaly we make `rel/config.exs` use it.

We also provide an update for OTP21+ for the args.

We keep the old vm.args for backward compatibility for project already using distillery.

There was a couple of formatting things added, i can get them out and open another PR if needed for them.
### Checklist

- [ x] Same for documentation, including moduledocs
- [x ] Tests were added or updated to cover changes
- [x ] Commits were squashed into a single coherent commit

### Licensing/Copyright


DianaOlympos:
Yeah that was the goal. Should make it easier for users to know where it come from and understand what they can do now :)

PR #571, opened 2018-10-15
fix io color formatting

Apelsinka223:
Hi, thanks for the package!
I found that debug output formatting is a bit broken

and also color extends to the following output.
I just added reset function at the and of the lines


bitwalker:
Thanks!

PR #578, opened 2018-10-25
Add note for ecto 3 users to bump the pool_size

LostKobrakai:
### Summary of changes

Ecto 3 requires a pool_size of 2. One for locking and one for running the migrations.

### Checklist

### Licensing/Copyright


EskiMag:
Wouldn't it be better to use `pool_size: 2` for all versions to prevent confusion?

LostKobrakai:
Maybe. I didn't change it because by now ecto 3 is still rc and people might still wonder why it's using not just one connection.

PR #591, opened 2018-11-21
Strip release in assembler instead of archiver.

mobileoverlord:
### Summary of changes

The rel directory should mimic what the actual output of the release. The setting `set strip_debug_info: true` makes modifications to the tar and not the release directory. This produces unexpected results for scenarios where the archive step is skipped. This PR moves this step to the assembly phase.


bitwalker:
Sorry for the delay man

PR #595, opened 2018-11-29
Fix formatting.

chenxsan:
The whole code block is missing now from the doc https://hexdocs.pm/distillery/guides/running_migrations.html:

image

Append a blank line fix the problem in my test:

image


bitwalker:
Thanks!

Issue #582, opened 2018-10-31
Unable to create PID file


dmytronasyrov:
1. `set erl_opts: "-kernel pidfile \'etc/server.pid\'"` in config.exs - doesn't work
2. `-kernel pidfile "'etc/server.pid'"` in vm.args - doesn't work
3. env variable `PIDFILE=etc/server.pid` - this works

```
{"could not start kernel pid",'Elixir.Mix.Releases.Runtime.Pidfile',{error {invalid_pidfile_config,'etc/server.pid'}}}
could not start kernel pid (Elixir.Mix.Releases.Runtime.Pidfile) ({error {invalid_pidfile_config,etc/server.pid}})
```

Trying to create PID file with kernel config but works only with env variable. Have tried different combinations of single and double quotes, but no luck. Any ideas or maybe a working snippet with a pidfile from vm.args or erl_opts?
Thank you.

ardhena:
I have similar issue with the PID file configuration - the one in vm.args file gave me the same error. I looked around in the code and manage to find that in that particular case, when the provided value has the format of `-kernel pidfile '"etc/server.pid"'` (notice the different order of `'` and `"`), the path variable in the code is a list  and in the code the pid file is only initiated if the path variable is a binary.
I opened a PR https://github.com/bitwalker/distillery/pull/600 with a fix for this, feel free to test if this works for you.

dmytronasyrov:
@ardhena yes, it works with your fix! @bitwalker should accept your PR. Thank you

PR #600, opened 2018-12-06
fix loading pidfile config from vm args

ardhena:
### Summary of changes

Connected with issue https://github.com/bitwalker/distillery/issues/582

**Current behaviour:**
When PID file configuration is loaded from the vm.args file in any of the following formats:
```
-kernel pidfile "'etc/server.pid'"
-kernel pidfile '"etc/server.pid"'
```
it creates the following error
```
{"could not start kernel pid",'Elixir.Mix.Releases.Runtime.Pidfile',{error {invalid_pidfile_config,'etc/server.pid'}}}
could not start kernel pid (Elixir.Mix.Releases.Runtime.Pidfile) ({error {invalid_pidfile_config,etc/server.pid}})
```
This prevents from having a working pidfile config in vm.args file

**Solution:**
After some debugging I found that in case of the second format (`'"etc/server.pid"'`) the path variable is a list, while in the code there is only check for binary https://github.com/bitwalker/distillery/blob/master/lib/mix/lib/releases/runtime/pidfile.ex#L34 It seems to me that maybe this value is an Erlang charlist, although I don't have much knowledge about Erlang, so I may be wrong.

Feel free to test this solution, it works for me for the format `'"etc/server.pid"'` in the vm.args file.

**NOTE:** If this PR will be approved, there should also be an update in the documentation, as now it shows a format of `"'etc/server.pid'"` (which is not working and is not fixed by this PR)

### Checklist

### Licensing/Copyright

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -46,7 +46,7 @@ provide a good overview of what has been addressed.
 - A new `config_providers` setting for defining which config providers to
   execute in a release (can be set in either `environment` or `release` blocks).
 - Support for writing a PID file to disk (useful for running under systemd in
-  particular). It is enabled with `-kernel pidfile "path/to/pidfile"` or by
+  particular). It is enabled with `-kernel pidfile '"path/to/pidfile"'` or by
   exporting `PIDFILE` in the environment. When enabled, the file is written to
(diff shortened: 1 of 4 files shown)


bitwalker:
Thanks for fixing this!

PR #614, opened 2018-12-18
Validate every app before adding to graph

fazibear:
There is a case when you want to add an application to release which don't exist.  For example umbrella app or an app from debian package.

Right now you'll get a nasty error:

```
  nil.name()
    (distillery) lib/mix/lib/releases/models/release.ex:532: Mix.Releases.Release.do_add_app/4
    (distillery) lib/mix/lib/releases/models/release.ex:547: Mix.Releases.Release.do_add_children/4
    (distillery) lib/mix/lib/releases/models/release.ex:507: Mix.Releases.Release.add_apps/3
    (distillery) lib/mix/lib/releases/models/release.ex:476: Mix.Releases.Release.apps/1
    (distillery) lib/mix/lib/releases/models/release.ex:330: Mix.Releases.Release.apply_configuration/3
    (distillery) lib/mix/lib/releases/assembler.ex:28: Mix.Releases.Assembler.pre_assemble/1
    (distillery) lib/mix/lib/releases/assembler.ex:46: Mix.Releases.Assembler.assemble/1
```

This PR fixes this case. Now you'll get:

```
==> Release failed, during .boot generation:
        Undefined applications: [backend]
```

### DIFF
--- a/lib/mix/lib/releases/models/release.ex
+++ b/lib/mix/lib/releases/models/release.ex
@@ -455,11 +455,8 @@ defmodule Mix.Releases.Release do
             name
         end
 
-      case Application.load(app_name) do
+      case validate_app(app_name) do
         :ok ->
-          :ok
-
-        {:error, {:already_loaded, _}} ->
           :ok
 
         {:error, reason} ->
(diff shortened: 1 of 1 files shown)


bitwalker:
Thanks!

PR #622, opened 2019-01-07
Windows: Remove erlsrv `-debugType new` option from install to stop unbounded logs

ellbee:
### Summary of changes

We have had extremely large log files generated on deploys to Windows servers. It seems like it is caused by the `-debugType new` option.

From the [erlsrv](http://erlang.org/doc/man/erlsrv.html) documentation:

>The new and reuse options might seem convenient in a production system, but consider that the logs grow indefinitely during the system lifetime and cannot be truncated, except if the service is restarted.

>In short, the DebugType is intended for debugging only. Logs during production are better produced with the standard Erlang logging facilities.

This pull request removes the `-debugType` flag entirely which has solved the problem for us. I am not sure what to do about the test section of the checklist, let me know if there is anything you require me to do here.

### Checklist

### Licensing/Copyright


bitwalker:
Thanks!

PR #624, opened 2019-01-07
Update mix phx.digest part

konstantinzolotarev:
### Summary of changes
Updated `cd` command because now if `PHOENIX_SUBDIR` need to be changed to something like `apps/web_api` it will fail on this step.

### Checklist


bitwalker:
Thanks!

PR #626, opened 2019-01-10
[#608] Windows fix for <app> console 

### LINKED ISSUE #608: [windows] <app> console starts incorrectly
### Version info
* erts 10.1
* elixir 1.7
* Distillery 2.0.12

### Steps to reproduce

Run `<app> console`

### Description of issue

When running `<app> console` I expect a local node to be started and the current console window to attach to that session. As it stands, an erlang window will appear momentarily before closing with the following appearing in the console

    PS C:\Users\hught\GuitarSniffer\Core\bin> .\guitar_sniffer_core.bat console

    Id     Name            PSJobTypeName   State         HasMoreData     Location             Command
    --     ----            -------------   -----         -----------     --------             -------
    1      post_start h... BackgroundJob   Running       True            localhost            ...

    PS C:\Users\hught\GuitarSniffer\Core\bin>

artman41:
### Summary of changes
*Fixes #608*
Modified console.ps1 to:
- execute the application inside the current powershell window
- hide the output of the `start-job` call
- close the elixir shell on `CTRL-C` or other breaking keycode

### Licensing/Copyright


priv/libexec/commands/win/console.ps1
```diff
@@ -18,6 +18,6 @@ $post_start = {
 }
 
 # Run post-start hooks asynchronously
-start-job -Name "post_start hooks" -ScriptBlock $post_start
+start-job -Name "post_start hooks" -ScriptBlock $post_start | out-null # hide the output from start-job
 
-& $werl @argv
+start-process "$erl" -ArgumentList "$argv" -Wait -NoNewWindow #execute the application in the current shell window
```
bitwalker: My impression was that `werl` is the only "true" Erlang shell, the command-line version is considerably simpler, missing many of the useful features - has that changed? I would expect someone would want the proper shell when attaching a console in production. Thoughts?

priv/libexec/commands/win/console.ps1
```diff
@@ -1,12 +1,12 @@
...
 $bin = whereis-erts-bin
-$werl = (join-path $bin werl)
+$erl = (join-path $bin "erl.exe") #get erl.exe as werl.exe will open a new window
 
 $boot = (join-path $Env:REL_DIR $Env:REL_NAME)
 
-$argv = @("-boot", $boot)
-$argv += @("-config", $Env:SYS_CONFIG_PATH)
-$argv += @("-args_file", $Env:VMARGS_PATH)
+$argv = @("-boot", "`"$boot`"")
```
bitwalker: Just for my own edification, could you explain the changes here? I'm going to merge them, I'd just like to know what I need to be doing differently or how I should be reasoning about quoting here; I'm working with PS Core for the most part, and I know there are differences with previous versions of PS, but from a maintenance standpoint I can't work with both, so Core is what I've stuck with, do these changes make any assumptions in that regard?

priv/libexec/commands/win/console.ps1
artman41: possibly I'm incorrect but I was under the assumption that the two were identical, in the case that they are not, it may be a good idea to have an optional config flag

priv/libexec/commands/win/console.ps1
artman41: so, the difference here is that we wrap the `$boot` in quotes so that paths containing a space work as intended, rather than acting as independent args 👍

PR #628, opened 2019-01-16
Fix Strip

michaelkschmidt:
This fixes an issue with using beam_lib to strip the release.  Some Elixir Apps (such as Cachex, see https://github.com/whitfin/cachex/issues/205)
need the 'Attr' chunk in order to work correctly, and it being removed.

The Stripper module is a direct copy/paste of the beam_lib functions


bitwalker:
I really don't want to reimplement the strip functionality from `beam_lib` in Distillery - my take is that if the stripped beams are _too_ stripped, then (for now) custom stripping needs to live in a plugin, and not in Distillery itself. 

The ideal approach is to open a PR with the OTP team to extend the `beam_lib` API to allow us to pass either a custom filtering function, or a list of chunks we want to keep above and beyond the required chunks; Distillery can then make use of this new API when it is available.

I appreciate the PR though, I agree that having a way to tweak the stripping would be a nice feature to have :)

bitwalker:
The strip option is _not_ provided by Distillery though - it is provided by `:systools`, which we expose via Distillery's config but explicitly _do not_ enable by default, specifically because it can cause issues. Stripping beams is only supposed to be done when the binary size of the extra chunks outweighs the value of those extra chunks, which for most applications is not usually worth the tradeoff.

In any case, yes, OTP PRs can take some time to trickle down to everyone, but those that truly need the functionality will upgrade as soon as its available, and that will be much much sooner than a year. In the meantime, the ability to strip beams manually (as you've done in this PR) remains an option via Distillery's plugin system. This PR is a few changes shy of being ready to be used as a `before_package` plugin, and could be open sourced as a stopgap measure for those who are in need of it.

michaelkschmidt:
I can migrate this to a plugin easily enough—it actually started out that way 😊

Any commercial Nerves project will likely need this feature. Stripping debug from embedded binaries is standard practice. Thus it’s suprising that this is broken

michaelkschmidt:
Here is the plugin
https://hex.pm/packages/strip_plugin

PR #679, opened 2019-06-11
Ensure RELEASE_MUTABLE_DIR exists before writing to it

mobileoverlord:
Fixes issue with trying to boot initial release. 

```
_build/dev/rel/example/bin/example console
cp: /Users/jschneck/Developer/nerves/shoehorn/example/_build/dev/rel/example/var/start_erl.data: No such file or directory
```


bitwalker:
Thanks!

PR #681, opened 2019-06-17
Fix REL_DIR in documentation

hrubi:
The `<vsn>` string was not seen as it was attempted to interpret it as a
tag. Also the variable does not hold the trailing slash.


bitwalker:
Thanks!

Issue #638, opened 2019-02-22
Dependency order seems wrong in the .rel file


jschniper:
### Steps to reproduce

I've created a repo at https://github.com/jschniper/distillery_order_ex with a sample application to reproduce the error. You might need to compile lager manually once in order for exometer to build. 

### Description of issue

This occurs with version 2.0.12 on Linux with Elixir 1.6.6.

The rel/config.exs is the generated version with no modifications.

Basically the core of the issue is that if I add the config flag for hackney to export metrics via exometer and build the release, the application won't start because hackney is starting prior to exometer. I've toyed with the order of the applications in the mix file as well as in the rel/config.exs file and nothing seems to fix the issue.

From the systools docs, "the applications are sorted according to the dependencies between the applications. Where there are no dependencies, the order in the .rel file is kept."

Basically, I would expect exometer to be higher in the .rel file because of either the rel/config.exs file or the mix.exs file but it's consistently lower. 

I'm wondering if the reverse sort that occurs at https://github.com/bitwalker/distillery/blob/master/lib/mix/lib/releases/models/release.ex#L483 might be causing this issue. I wasn't sure why the reverse sorting was happening and looking back at the commit (https://github.com/bitwalker/distillery/commit/ab8d308abfbfcd07a5df5ede2bebbf05ae2e65a0) it looked as though it might be code that was left over. This is all speculative, of course, and I can't wait to find out why I'm wrong but I've tried everything I could think of at this point to no avail.

Let me know if you need more information.

bitwalker:
The ultimate decider of this stuff is `systools`, so we have pretty minimal influence on how dependencies get sorted. That said, I'm open to removing the `reverse` call there, and seeing if it fixes the issue you are seeing. If so, then I'll merge that fix into master. Unfortunately I don't have the time to dig deeper myself, so if you can give that a shot and let me know the result, I can take it the last mile.

jschniper:
Removing that reverse did fix my issue. Do you want me to get a pull request together for this?

bitwalker:
Yes please!

PR #689, opened 2019-06-26
Remove the reverse on dependency order (see: bitwalker/distillery#638)

jschniper:
### Summary of changes

Removes the reverse on the dependency list in order to ensure the order is correct when an application starts.

### Licensing/Copyright


bitwalker:
Thanks!

PR #703, opened 2019-08-16
Always update start_erl.data

### LINKED ISSUE #693: Discrepancies between two versions of the start_erl.data
### Steps to reproduce
Not 100% sure since our deployment process is complex. 

Just a brief list of the steps required to produce the issue.
I'm seeing a discrepancy between the two `start_erl.data` files. 

`/opt/my_app/releases/start_erl.data` shows correctly 
`/opt/insights/var/start_erl.data` has a git commit sha that is a few commits back. 

I found this as a result of trying to run a migration script that called on 
`/opt/myapp/bin/myapp_rc_exec.sh`

I have found this issue only after this commit #670 
https://github.com/bitwalker/distillery/commit/4ebfe9a6464d981b77008e083d258b4b720c6838#diff-bb0b28dac91fa3fff22ffea720be6334

Any advice as to the difference between the two versions of `start_erl.data` and what is the reasoning  behind the code below is would be greatly appreciated. 

```
RELEASE_MUTABLE_DIR="${RELEASE_MUTABLE_DIR:-"${RELEASE_ROOT_DIR}/var"}"
if [ ! -f "${START_ERL_DATA}" ]; then
    mkdir -p $RELEASE_MUTABLE_DIR
    cp "${RELEASES_DIR}/start_erl.data" "${START_ERL_DATA}"
fi
```

adkron:
The start_erl.data file is not getting updated to the newest version.
This means that when starting the application an old version is
actually starting up.

Closes #693 

Amos King @adkron <amos@binarynoggin.com>

### DIFF
--- a/priv/templates/release_rc_exec.eex
+++ b/priv/templates/release_rc_exec.eex
@@ -11,10 +11,9 @@ RELEASE_MUTABLE_DIR="${RELEASE_MUTABLE_DIR:-"${RELEASE_ROOT_DIR}/var"}"
 START_ERL_DATA="${RELEASE_MUTABLE_DIR}/start_erl.data"
 REL_NAME="${REL_NAME:-<%= release_name %>}"
 
-if [ ! -f "${START_ERL_DATA}" ]; then
-    mkdir -p $RELEASE_MUTABLE_DIR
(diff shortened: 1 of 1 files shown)


bitwalker:
Closing in favor of #718. Thanks for the PR though!

PR #708, opened 2019-09-17
Add more information about upgrading from 2.0 to 2.1 

JonRowe:
Hi, I struggled to find this information easily as the change log was a bit hard for me to find (which it turns out is just pure selective reading, I didn't think it'd be in the docs) but I was surprised to not find the information in upgrading, so this PR adds a section about upgrading to 2.1

Note I'm happy to remove the renaming if you feel its inappropriate, and I'm happy to drop the commit removing the note about semver but it removes discussion about the fact that semver was broken for this change. See https://semver.org/spec/v2.0.0.html :

> **What do I do if I accidentally release a backwards incompatible change as a minor version?**
>As soon as you realize that you’ve broken the Semantic Versioning spec, fix the problem and release a new minor version that corrects the problem and restores backwards compatibility. Even under this circumstance, it is unacceptable to modify versioned releases. If it’s appropriate, document the offending version and inform your users of the problem so that they are aware of the offending version.

### Licensing/Copyright


JonRowe:
Closed as this PR has gone over a year without any input, I assume its unwanted.

PR #718, opened 2019-11-22
fix for start_erl.data - old release getting started after deployment

delta1:
### Summary of changes 

This is a simple change to fix the issue in #693 - where the start_erl.data file in releases/ does not get copied into the mutable dir when the directory already exists. 

It is a slightly better fix than #703 since it checks if the start_erl.data file in releases dir is newer than the var version first, instead of blindly copying it every single time. 

### Checklist

- [ n/a] New functions have typespecs, changed functions were updated
- [ n/a] Same for documentation, including moduledocs
- [?] Tests were added or updated to cover changes - **not sure how to test this change, please advise**

### Licensing/Copyright

### DIFF
--- a/priv/templates/release_rc_exec.eex
+++ b/priv/templates/release_rc_exec.eex
@@ -11,8 +11,10 @@ RELEASE_MUTABLE_DIR="${RELEASE_MUTABLE_DIR:-"${RELEASE_ROOT_DIR}/var"}"
 START_ERL_DATA="${RELEASE_MUTABLE_DIR}/start_erl.data"
 REL_NAME="${REL_NAME:-<%= release_name %>}"
 
-if [ ! -f "${START_ERL_DATA}" ]; then
+if [ ! -d "${RELEASE_MUTABLE_DIR}" ]; then
     mkdir -p $RELEASE_MUTABLE_DIR
+fi
+if [[ "${RELEASES_DIR}/start_erl.data" -nt "${START_ERL_DATA}" ]]; then
     cp "${RELEASES_DIR}/start_erl.data" "${START_ERL_DATA}"
 fi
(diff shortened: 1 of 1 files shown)


delta1:
@bitwalker ?

aselder:
@bitwalker Can we get this merged? This is a fix for a serious issue.

tmartin314:
As a temp workaround, I added this hook to the `.deliver/config`.  
```
post_extract_release_archive() {
  status "Copying new release version start_erl.data file"
  __sync_remote "
    [ -f ~/.profile ] && source ~/.profile
    set -e
    cp <absolutepath>/releases/start_erl.dat <absolutepath>/var/start_erl.data' $SILENCE
  "
}
```
Thanks for the PR @delta1

delta1:
@tmartin314 your cp command has `start_erl.dat` as the first argument instead of `start_erl.data` ;)

nocivus:
If you happen to be using bootleg instead, here's a similar fix: https://github.com/labzero/bootleg/issues/313

Scarysize:
Nice fix. Would be awesome to have a new patch release for the distillery package!

Ninigi:
@bitwalker thank you so much for making most of Elixir deployment a breeze. This PR has been merged for a few months now, but I still have to deal with this issue on every single project. Can we get a hex release please? I would prefer broken distillery (which I know might be broken) over playing Groundhog Day. (and no, I do not want to install from master, that is even worse than dealing with the same problem over and over again.)

delta1:
@Ninigi save your sanity by just forking and then referencing your fork in `mix.exs` 

```elixir
{:edeliver, github: "user/edeliver", branch: "master"} # or ref: or tag:
``` 

`https://hexdocs.pm/mix/Mix.Tasks.Deps.html`

PR #723, opened 2020-01-19
Note inability to remove pifile on SIGKILL

brucepom:
I've run into issues where Systemd was unable to detect that my app had crashed because the app failed to remove its pidfile. This happened to me because the Beam was killed using a `SIGKILL` by Ubuntu's "Out Of Memory Killer". Under these conditions, the app is killed suddenly and wasn't able to remove it's pidfile. I am using the [pid_file 0.1.0](https://github.com/OvermindDL1/pid_file) to manage pidfiles.


bitwalker:
Thanks!

PR #724, opened 2020-01-19
Note about enabling service to start on boot

brucepom:
Useful note, as it's not obvious that you need to enable the service if you want it to be restarted when the server is restarted.


bitwalker:
Thanks!

